PR #46, opened 2018-10-16
Add compat module for py2 py3

maxkrivich:
* Add restrictions on python version to setup.py;
* Add compat.py to handle all import compatibility issues between python versions.


annatisch:
@maxkrivich - This is awesome - thank you so much for your contribution :)

Issue #53, opened 2018-12-07
Async send stuck


zach-b:
I am using the client to send messages to the azure service bus. I noticed that some messages were never delivered.

The SendClientAsync.send_message_async method is stuck in the loop 
```
while any([m for m in pending_batch if m.state not in constants.DONE_STATES]):    
    await self.do_work_async()
```

The message is in the state <MessageState.WaitingToBeSent: 0> and is not in self._pending_messages, so I am thinking that it might have never been added to the list due to some concurrency issue ?

annatisch:
Thanks @zach-b!
Interesting that you're not seeing the message in _pending_messages - could I please see a code snippet that you're using to send?

A couple of additional questions:
- Which OS and Python version are running on?
- How much data is in the message that is hanging?

We have a current issue in Event Hubs for messages with content greater than ~128kb getting stuck which may be related:
https://github.com/Azure/azure-event-hubs-python/issues/76

zach-b:
Thanks for your answer

The message is sent using the following method
```
    async def send(self, msg, ttl=None):
            ttl_header = None
            if ttl is not None:
                ttl_header = MessageHeader()
                ttl_header.time_to_live = ttl * 1000
            umsg = UamqpMessage(
                body=msg.body,
                application_properties=msg.headers,
                header=ttl_header
            )
            await self._sender.send_message_async(umsg)
```
The OS is Debian 8, I did not check the content of the messages yet, I'll make sure to do it if the bug occurs again tonight (or else on monday), but the messages should be rather small with a few headers and no body. I'll try to isolate and reproduce the bug on monday

zach-b:
So I investigated a bit further. and the content of the messages is indeed as described above (nothing out of the ordinary...).
I was not able to reproduce the bug on a simple case for the moment, my guess is that it is due to concurrency issue with the pending messages list, but I cannot know for sure. I'll try again in the following days

annatisch:
Thanks for the follow-up @zach-b - yes I agree that seems to be the likely cause. Especially given that it's so intermittent. I will investigate.

zach-b:
I tried adding a lock to the async methods modyfing the pending messages list, I'm letting it run for some time to see if any message still gets stuck (it usually happened a few times per hour). I'll make a PR if it's successful

annatisch:
Oh you're a legend! Thank you so much for your continued investigations! I look forward to hearing how it goes :)

zach-b:
[...]

PR #54, opened 2018-12-11
Prevent concurrent modifications of SendClientAsync._pending_messages

zach-b:
Fix issue #53

### DIFF
--- a/uamqp/async_ops/client_async.py
+++ b/uamqp/async_ops/client_async.py
@@ -630,7 +630,8 @@ class SendClientAsync(client.SendClient, AMQPClientAsync):
         """
         await self.open_async()
         try:
-            messages = self._pending_messages[:]
+            async with self._pending_messages_lock:
+                messages = self._pending_messages[:]
             await self.wait_async()
             results = [m.state for m in messages]
             return results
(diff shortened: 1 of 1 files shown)


annatisch:
These changes look good thanks @zach-b!
I feel we could benefit from also applying the lock in [line 633](https://github.com/Azure/azure-uamqp-python/blob/41a27f421f6d2bc9bc624d10ed84258c49120251/uamqp/async_ops/client_async.py#L633): `messages = self._pending_messages[:]` to prevent pending_messages being snapshot during the filtering operation.

Before releasing the fix, I may need to do some additional testing here with regards to:
- The `def messages_pending()` function that is currently only implemented synchronously. Potentially this can be called while the filtering operation is in progress, so we may need to implement an asynchronous version of this function that waits for the lock before reporting whether messages are pending.
- Adding the equivalent `threading.Lock` to the synchronous client, as it's surely vulnerable to the same issue

zach-b:
@annatisch I applied the lock in line 633 as well

Regarding the messages_pending() property, it would require creating a messages_pending_async() method, however there might be a risk that developers would use the message_pending property anyway instead ?

PR #58, opened 2019-03-05
Break out openssl job

bsiegel:


annatisch:
Maybe now would be a good time to see if we can get the MacOS builds running without openssl....
Might save us some pain :)

bsiegel:
> Maybe now would be a good time to see if we can get the MacOS builds running without openssl....
> Might save us some pain :)

I'm 1000% in favor of that, but this PR won't really add much complexity. Just having the OpenSSL libs build nightly so that mac PR builds aren't so slow.

PR #62, opened 2019-03-21
Missing application_properties in BatchMessage fixes

yunhaoling:
https://github.com/Azure/azure-event-hubs-python/issues/97

### DIFF
--- a/uamqp/message.py
+++ b/uamqp/message.py
@@ -464,19 +464,21 @@ class BatchMessage(Message):
             body_size = 0
             try:
                 for data in self._body_gen:
-                    message_segment = []
-                    if isinstance(data, six.text_type):
-                        data = data.encode(self._encoding)
-                    batch_data = c_uamqp.create_data(data)
-                    c_uamqp.enocde_batch_value(batch_data, message_segment)
-                    combined = b"".join(message_segment)
-                    body_size += len(combined)
+                    message_bytes = None
+                    if not isinstance(data, Message):  # raw data
+                        wrap_message = Message(body=data, application_properties=self.application_properties)
+                        message_bytes = wrap_message.encode_message()
+                    else:
+                        if not data.application_properties:  # Message without app_prop
+                            data.application_properties = self.application_properties
+                        message_bytes = data.encode_message()
+                    body_size += len(message_bytes)
                     if (body_size + message_size) > self.max_message_length:
                         new_message.on_send_complete = self.on_send_complete
                         yield new_message
                         raise StopIteration()
                     else:
-                        new_message._body.append(combined)  # pylint: disable=protected-access
+                        new_message._body.append(message_bytes)  # pylint: disable=protected-access
             except StopIteration:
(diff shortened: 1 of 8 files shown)


uamqp/message.py
```diff
@@ -458,25 +458,32 @@ def _multi_message_generator(self):
...
                 for data in self._body_gen:
-                    message_segment = []
-                    if isinstance(data, six.text_type):
-                        data = data.encode(self._encoding)
-                    batch_data = c_uamqp.create_data(data)
-                    c_uamqp.enocde_batch_value(batch_data, message_segment)
-                    combined = b"".join(message_segment)
-                    body_size += len(combined)
+                    message_bytes = None
+                    if not isinstance(data, Message):  # raw data
```
annatisch: After giving it some thought - I think this type check could be easier to adapt to the EH SDK if we do it like this:
```python
try:
    if not data.application_properties:  # Message-like object
        data.application_properties = self.application_properties
    message_bytes = data.encode_message()
except AttributeError:  # raw data
    wrap_message = Message(body=data, application_properties=self.application_properties)
    message_bytes = wrap_message.encode_message()
```
This means we could consume any "message-like" object, so long as it has an "application_properties" attribute and an "encode_message" function. This should make it easier to adapt the EH SDK, as an `Event` object already has the `application_properties` attribute, we just need to give it an `encode_message` function and we should be able to use it directly.

uamqp/message.py
```diff
@@ -501,19 +508,20 @@ def gather(self):
...
-            message_segment = []
-            if isinstance(data, six.text_type):
-                data = data.encode(self._encoding)
-            batch_data = c_uamqp.create_data(data)
-            c_uamqp.enocde_batch_value(batch_data, message_segment)
-            combined = b"".join(message_segment)
-            body_size += len(combined)
-
+            message_bytes = None
+            if not isinstance(data, Message):  # raw data
```
annatisch: Same type-check comment as above.

uamqp/message.py
yunhaoling: Great idea! Get what you mean and it reminds me of duck typing.
This is undoubtedly a better implementation. I think what we should do after taking this implementation is to provide more detailed documentations about batching in EventData.

annatisch:
Hi @yunhaoling - I updated pylint as it was throwing errors due a version conflict, however this has resulted in some new formatting errors. Could you also fix these? There's not too many:
https://travis-ci.org/Azure/azure-uamqp-python/jobs/511066675

yunhaoling:
@annatisch Sure, I'll investigate into that.

annatisch:
Hi @yunhaoling - I would revert your last two commits and simply remove Python 3.4 from the travis config:
https://github.com/Azure/azure-uamqp-python/blob/master/.travis.yml#L14

We don't plan to support Python 3.4 from the next release anyway :)

yunhaoling:
Hey @annatisch 
I've done the rollback and removed python 3.4 from travis config.

PR #71, opened 2019-05-06
Adding livetest yaml

scbedd:
Note that this doesn't change anything for builds triggered by `PR` or any external source. 

Ensuring this works, expect IOTHub and RabbitMQ env sets in further commits.

### DIFF
--- a/.azure-pipelines/client.test.live.yml
+++ b/.azure-pipelines/client.test.live.yml
@@ -9,42 +9,125 @@ variables:
   PythonVersion38: '3.8.0a2'
 
 jobs:
-  - job: 'sdist'
-    displayName: 'Source Distribution'
+  - job: 'Windows'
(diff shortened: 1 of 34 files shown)


annatisch:
Thanks!
You can ignore the RabbitMQ env vars for now. These would be set by running the tests in a container rather than any need for keyvault secrets. And they're currently not passing and ignored anyway....

scbedd:
Awesome, then it's just IOTHub remaining once I confirm this is working.

scbedd:
DevOps simply doesn't support variable `strategy` settings. I can't simply set maxParallel to 1 in the case of being inside the `internal` project or the like. As a result, I'm simply splitting off the definition, removing the `build` and `manylinux` steps, and adding environment variables to enable livetests.

PR #74, opened 2019-05-13
Event Hub Track 2 Support

yunhaoling:
- [x] WebSocket Support https://github.com/Azure/azure-sdk-for-python/issues/5071
 - [x] OAuth Support https://github.com/azure/azure-sdk-for-python/issues/5097


uamqp/constants.py
```diff
@@ -18,6 +19,8 @@
 OPERATION = b"operation"
 READ_OPERATION = b"READ"
 MGMT_TARGET = b"$management"
+EVENT_HUB_SCOPE = ['https://eventhubs.azure.net//.default']
+SERVICE_BUS_SCOPE = ['https://servicebus.azure.net//.default']
```
annatisch: Please remove these - we shouldn't keep references to Azure services in uAMQP. These should be passed through as values from EventHub or ServiceBus SDKs

yunhaoling: @annatisch Gotcha, just removed 😃

yunhaoling: I found Charles has updated the azure-identity, I need to do some modification to the JWTTokenAsync. I'll update you today when this pr is ready to merge.

PR #75, opened 2019-05-23
Event Hub Track 2 Support

yunhaoling:
@annatisch 

Hey Anna, please review my PR, thanks! 

- [x] Asyn JWTToken support 
    - Related issue: https://github.com/Azure/azure-sdk-for-python/issues/5361
 - [x] Update MAX_MESSAGE_LENGTH_BYTES to 1MB
    - Related issue: https://github.com/Azure/azure-sdk-for-python/issues/5286


uamqp/authentication/cbs_auth_async.py
```diff
@@ -221,3 +235,56 @@ class JWTTokenAsync(JWTTokenAuth, CBSAsyncAuthMixin):
...
+            raise ValueError("Must specify either 'expires_at' or 'expires_in'.")
+        if not expires_at:
+            self.expires_in = expires_in
+            self.expires_at = time.time() + expires_in.seconds
+        else:
+            self.expires_at = expires_at
+            expires_in = expires_at - time.time()
+            if expires_in < 1:
+                raise ValueError("Token has already expired.")
+            self.expires_in = datetime.timedelta(seconds=expires_in)
```
annatisch: This section of code is duplicated 4 times between all 4 token type - we should move it to a method on the CBSAuthMixin:
```python
class CBSAuthMixin:
    def _set_expiry(self, expires_at, expires_in):
        if not expires_at and not expires_in:
            raise ValueError("Must specify either 'expires_at' or 'expires_in'.")
        if not expires_at:
            expires_at = time.time() + expires_in.seconds
        else:
            expires_in_seconds = expires_at - time.time()
        if expires_in_seconds < 1:
            raise ValueError("Token has already expired.")
            expires_in = datetime.timedelta(seconds=expires_in_seconds)
        return expires_at, expires_in

Token(CBSAuthMixin):
    def __init__(self, ...):
        ....
        self.expires_at, self.expires_in = self._set_expiry(expires_at, expires_in)
        ...
```

yunhaoling:
Hey @annatisch , code updated!😃

uamqp/authentication/cbs_auth_async.py
```diff
@@ -221,3 +245,40 @@ class JWTTokenAsync(JWTTokenAuth, CBSAsyncAuthMixin):
...
+        self._retry_policy = retry_policy
+        self._encoding = encoding
+        self.uri = uri
+        parsed = compat.urlparse(uri)  # pylint: disable=no-member
+
+        self.cert_file = verify
+        self.hostname = parsed.hostname.encode(self._encoding)
+
+        if not is_coroutine(get_token):
+            raise ValueError("get_token must be a coroutine function")
```
annatisch: You don't need to raise a ValueError here because the `is_coroutine` function already raises it

uamqp/message.py
```diff
@@ -517,9 +517,7 @@ def gather(self):
                 message_bytes = wrap_message.encode_message()
             body_size += len(message_bytes)
             if (body_size + message_size) > self.max_message_length:
-                raise ValueError(
-                    "Data set too large for a single message."
-                    "Set multi_messages to True to split data across multiple messages.")
+                raise errors.ClientMessageError("Data set too large for a single message.")
```
annatisch: It occurs to me that this is actually a breaking change. Could we do an intermediate fix for this that's non breaking - then adjust it later next time there's a major version bump?
I'm thinking, in errors.py add:
```
class MessageContentTooLarge(ValueError):
    def __init__(self):
        message = "Data set too large for a single message."
        super(ValueError, self).__init__(message)
```
Then use this error in message.py

PR #78, opened 2019-06-06
Fix livetest error and compatibility with EventHub Track1

yunhaoling:


uamqp/async_ops/client_async.py
```diff
@@ -105,7 +105,16 @@ def __init__(
...
+
+        if loop:
+            self.loop = loop
+        else:
+            try:
+                if not self.loop:  # from sub class instance
+                    self.loop = get_running_loop()
+            except AttributeError:
+                self.loop = get_running_loop()
+
```
annatisch: I think this code could be simplified with:
```python
self.loop = loop
if not self.loop:
    self.loop = get_running_loop()
```

yunhaoling: I'm afraid we can't simplify it this way. 

For an instance of SendClientAsync which extends from AMQPClientAsync, it first initializes its own self.loop member in the SendClientAsync constructor and then it will call its fathers __init__ function. If we write self.loop = loop in AMQPClientAsync constructor, the loop member instance of the SendClientAsync will be overwritten to None (The default loop parameter is None in AMQPClientAsync __init__) which leads to a bug.

Let me know if you want a more detailed explanation.

PR #88, opened 2019-08-19
Share CBS Session

yunhaoling:
Initialize the `_received_messages` when in the `__init__` function of the AMQP client so that received messages won't be simply dropped during the authentication process of multiple clients.

-----

@annatisch 
I'm wondering whether the `self._streaming_receive` should be thread-safe or not.

The situation where thread-unsafe may happen to the `_streaming_receive` variable is that both functions `receive_message_batch` and `receive_messages` are called simultaneously.

But I don't think our SDK support such kind of operation, do I miss something?

### DIFF
--- a/samples/asynctests/test_azure_event_hubs_receive_async.py
+++ b/samples/asynctests/test_azure_event_hubs_receive_async.py
@@ -131,7 +131,6 @@ async def test_event_hubs_batch_receive_async(live_eventhub_config):
 
 @pytest.mark.asyncio
 async def test_event_hubs_shared_connection_async(live_eventhub_config):
-    pytest.skip("Unstable on OSX and Linux - need to fix")  # TODO
     uri = "sb://{}/{}".format(live_eventhub_config['hostname'], live_eventhub_config['event_hub'])
     sas_auth = authentication.SASTokenAsync.from_shared_access_key(
         uri, live_eventhub_config['key_name'], live_eventhub_config['access_key'])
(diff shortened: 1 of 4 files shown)


yunhaoling:
The function `release_async` in connection_async.py isn't a co-routine function.
The name is a bit misleading as we may think `release-async` is an async function and call it by `await release_async()`.

uamqp/client.py
```diff
@@ -1097,7 +1102,7 @@ def receive_messages_iter(self, on_message_received=None):
         :type on_message_received: callable[~uamqp.message.Message]
         """
         self._message_received_callback = on_message_received
-        self._received_messages = compat.queue.Queue()
+        self._received_messages = self._received_messages or compat.queue.Queue()
```
annatisch: Is this assignment statement still necessary? `self._received_messages` should never be `None` now right?

uamqp/async_ops/client_async.py
```diff
@@ -945,7 +950,7 @@ def receive_messages_iter_async(self, on_message_received=None):
         :rtype: Generator[~uamqp.message.Message]
         """
         self._message_received_callback = on_message_received
-        self._received_messages = queue.Queue()
+        self._received_messages = self._received_messages or compat.queue.Queue()
```
annatisch: Is this assignment statement still necessary? `self._received_messages` should never be `None` now right?

samples/asynctests/test_azure_event_hubs_receive_async.py
```diff
@@ -131,7 +131,6 @@ def on_message_received(message):
 
 @pytest.mark.asyncio
 async def test_event_hubs_shared_connection_async(live_eventhub_config):
-    pytest.skip("Unstable on OSX and Linux - need to fix")  # TODO
```
annatisch: Awesome that these are now resolved :)
Once this is merged I think we have an issue for these that we can now close?

PR #89, opened 2019-08-30
Share CBS Session

yunhaoling:
- [x] fix bug of sharing cbs session

### DIFF
--- a/uamqp/async_ops/client_async.py
+++ b/uamqp/async_ops/client_async.py
@@ -217,47 +217,49 @@ class AMQPClientAsync(client.AMQPClient):
         # pylint: disable=protected-access
         if self._session:
             return  # already open
-        if connection:
-            _logger.info("Using existing connection.")
-            self._auth = connection.auth
-            self._ext_connection = True
-            await connection.lock_async()
-        self._connection = connection or self.connection_type(
-            self._hostname,
-            self._auth,
-            container_id=self._name,
-            max_frame_size=self._max_frame_size,
-            channel_max=self._channel_max,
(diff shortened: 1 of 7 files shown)


uamqp/client.py
```diff
@@ -871,7 +874,8 @@ def __init__(
         self._last_activity_timestamp = None
         self._was_message_received = False
         self._message_received_callback = None
-        self._received_messages = None
+        self._streaming_receive = False
+        self._received_messages = compat.queue.Queue()
```
annatisch: These two lines don't seem to be in `client_async.close_async()`?

yunhaoling: you mean resetting these two variables when a client is closed? but we don't support open-close-open..

uamqp/client.py
```diff
@@ -270,6 +271,8 @@ def open(self, connection=None):
         if self._keep_alive_interval:
             self._keep_alive_thread = threading.Thread(target=self._keep_alive)
             self._keep_alive_thread.start()
+        if self._ext_connection:
+            connection.release()
```
annatisch: I think this should go in a `try/finally` statement

uamqp/async_ops/client_async.py
```diff
@@ -256,6 +256,8 @@ def __init__(
                 loop=self.loop)
         if self._keep_alive_interval:
             self._keep_alive_thread = asyncio.ensure_future(self._keep_alive_async(), loop=self.loop)
+        if self._ext_connection:
+            connection.release_async()
```
annatisch: Same comment re `try/finally`

PR #90, opened 2019-08-30
Close the amqp client when timeout

yunhaoling:


uamqp/async_ops/client_async.py
```diff
@@ -914,22 +914,25 @@ def __init__(
...
+                    if self._received_messages.qsize() > 0 and received == 0:
+                        # No new messages arrived, but we have some - so return what we have.
+                        expired = True
+                        break
+                while not self._received_messages.empty() and len(batch) < max_batch_size:
+                    batch.append(self._received_messages.get())
+                    self._received_messages.task_done()
+        finally:
+            if self._shutdown:
+                self.close()
```
annatisch: This should be `await self.close_async()` I think

yunhaoling: updated:😉

yunhaoling:
close this for now as it needs further investigation

PR #92, opened 2019-09-06
Fix websocket support.

skoop22:
amqp over Websocket does not work when only http_proxy is defined.

The documentation states of the AMQPAuth class states that when http_proxy is defined the transport_type is implicitly set to uamqp.TransportType.AmqpOverWebsocket. This was not the case...


annatisch:
This is awesome - thanks @skoop22 so much for the fix :)

PR #94, opened 2019-09-19
Support message delivery tag

annatisch:
First pass at exposing the binary data provided in the message delivery tag.
This is used by ServiceBus to provide the lock token.

### DIFF
--- a/samples/test_azure_event_hubs_receive.py
+++ b/samples/test_azure_event_hubs_receive.py
@@ -123,7 +123,9 @@ def test_event_hubs_client_receive_sync(live_eventhub_config):
             assert len(batch) <= 10
             for message in batch:
                 annotations = message.annotations
-                log.info("Sequence Number: {}".format(annotations.get(b'x-opt-sequence-number')))
+                log.info("Sequence Number: {}, Delivery tag: {}".format(
+                    annotations.get(b'x-opt-sequence-number'),
+                    message.delivery_tag))
             batch = receive_client.receive_message_batch(max_batch_size=10)
     log.info("Finished receiving")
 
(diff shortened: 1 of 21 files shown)


yunhaoling:
Hello @annatisch 

Would it be better to let the message receiver to assign `delivery-tag` of each `Transfer` frame?

- The `delivery-tag` seems to be related to each `Transfer` frame (which is equal to a message I think).
- If we only set the latest `delivery-tag` on the `Link`, then what about the following situation (I'm not sure whether it's the use case for ServiceBus):
    - Step1: receive the 1st msg
    - Step2: receive the 2rd msg
    - Step3: Now I want to lock the 1st msg (but the last `delivery-tag` of the receiver is for the 2rd message)

I found that the link will [pass the handling of Transfer to the receiver](https://github.com/Azure/azure-uamqp-python/blob/e40e6555ed9a28835ae7390e593c0b2b0800882d/src/vendor/azure-uamqp-c/src/link.c#L483).
How about handling `Transfer` and assigning the `delivery-tag` to the message created in [the on_transfer_received of the message_receiver](https://github.com/Azure/azure-uamqp-python/blob/e40e6555ed9a28835ae7390e593c0b2b0800882d/src/vendor/azure-uamqp-c/src/message_receiver.c#L223)?

annatisch:
Thanks @yunhaoling - yeah good call - I will refactor.

luipir:
only @yunhaoling as possible reviewer?

annatisch:
Hi @yunhaoling - I have reverted my debugging and this is now ready to review.

annatisch:
@luipir  - if you look at the builds you can download one of the test wheels to test out locally if you like:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=118233

Just select your chosen platform flavor from the "Artifacts" drop down.

annatisch:
Closing this in favour of PR #96 which includes some additional fixes.

luipir:
@annatisch sorry I hadn't time to test yet ... I'll do (the new PR) as soon as I finish other code. Tnx for the (possible) fix.

PR #95, opened 2019-09-25
Runtime metric

yunhaoling:


src/link.pyx
```diff
@@ -136,6 +136,20 @@ cdef class cLink(StructBase):
             self._value_error()
         return value
 
+    @property
+    def desired_properties(self):
+        cdef c_amqpvalue.AMQP_VALUE value
+        if c_link.link_get_desired_capabilities(self._c_value, &value) != 0:
+            self._value_error()
+        return value
```
annatisch: Do we need to call `value_factory` of the output to convert it to Python data?

src/link.pyx
```diff
@@ -136,6 +136,20 @@ cdef class cLink(StructBase):
...
+
+    @property
+    def offered_properties(self):
+        cdef c_amqpvalue.AMQP_VALUE value
+        if c_link.link_get_offered_capabilities(self._c_value, &value) != 0:
+            self._value_error()
+        return value
```
annatisch: Do we need to call `value_factory` of the output to convert it to Python data?

src/vendor/azure-uamqp-c/src/link.c
```diff
@@ -808,6 +835,20 @@ LINK_HANDLE link_create_from_endpoint(SESSION_HANDLE session, LINK_ENDPOINT_HAND
...
+		// if (amqpvalue_add_array_item(desired_capabilities_array, one_symbol) != 0) {
+		// 	LogError("Cannot create tick counter for link");
+		// 	free(result);
+		// 	result = NULL;
+		// 	return result;
+		// }
+
+		// result->desired_capabilities = desired_capabilities_array;
+        result->desired_capabilities = NULL;
+		result->offered_capabilities = NULL;
```
annatisch: Indentation

src/vendor/azure-uamqp-c/src/link.c
```diff
@@ -725,12 +736,28 @@ LINK_HANDLE link_create(SESSION_HANDLE session, const char* name, role role, AMQ
...
+		// 	LogError("Cannot create tick counter for link");
+		// 	free(result);
+		// 	result = NULL;
+		// 	return result;
+		// }
+
+		// result->desired_capabilities = desired_capabilities_array;
+
+        result->desired_capabilities = NULL;
+		result->offered_capabilities = NULL;
```
annatisch: Indentation

src/vendor/azure-uamqp-c/src/link.c
```diff
@@ -1083,6 +1124,84 @@ int link_get_peer_max_message_size(LINK_HANDLE link, uint64_t* peer_max_message_
...
+    }
+    else if ((link->link_state != LINK_STATE_ATTACHED) &&
+    (link->link_state != LINK_STATE_HALF_ATTACHED_ATTACH_RECEIVED))
+    {
+                LogError("Attempting to read offered capabilities before it was received");
+        result = __FAILURE__;
+    }
+    else
+    {
+        *offered_capabilities = link->offered_capabilities;
```
annatisch: We should do an `amqpvalue_clone` here

src/vendor/azure-uamqp-c/src/link.c
```diff
@@ -1083,6 +1124,84 @@ int link_get_peer_max_message_size(LINK_HANDLE link, uint64_t* peer_max_message_
...
+    int result;
+
+    if (link == NULL)
+    {
+        LogError("NULL link");
+        result = __FAILURE__;
+    }
+    else
+    {
+        link->offered_capabilities = offered_capabilities;
```
annatisch: We need to do `amqpvalue_clone` on the input before assigning it.

src/vendor/azure-uamqp-c/src/link.c
```diff
@@ -1083,6 +1124,84 @@ int link_get_peer_max_message_size(LINK_HANDLE link, uint64_t* peer_max_message_
...
+    if((link == NULL) ||
+        (desired_capabilities == NULL))
+    {
+        LogError("Bad arguments: link = %p, desired_capabilities = %p",
+            link, desired_capabilities);
+        result = __FAILURE__;
+    }
+    else
+    {
+        *desired_capabilities = link->desired_capabilities;
```
annatisch: Same comment re `amqpvalue_clone`

src/vendor/azure-uamqp-c/src/link.c
```diff
@@ -1083,6 +1124,84 @@ int link_get_peer_max_message_size(LINK_HANDLE link, uint64_t* peer_max_message_
...
+    int result;
+
+    if (link == NULL)
+    {
+        LogError("NULL link");
+        result = __FAILURE__;
+    }
+    else
+    {
+        link->desired_capabilities = desired_capabilities;
```
annatisch: Same comment re `amqpvalue_clone`

src/vendor/azure-uamqp-c/src/link.c
```diff
@@ -278,6 +280,15 @@ static int send_attach(LINK_INSTANCE* link, const char* name, handle handle, rol
...
         }
+        
+        if (link->desired_capabilities != NULL)
+        {
+            if(attach_set_desired_capabilities(attach, link->desired_capabilities) != 0)
+            {
+                LogError("Cannot set attach desired capabilities");
+                result = __FAILURE__;
+            }
+        }
```
annatisch: I see we're setting the desired capabilities here, but we're not extracting the offered capabilities from the received attach frame?

src/vendor/inc/c_link.pxd
```diff
@@ -52,6 +52,10 @@ cdef extern from "azure_uamqp_c/link.h":
     int link_get_max_message_size(LINK_HANDLE link, stdint.uint64_t* max_message_size)
     int link_get_peer_max_message_size(LINK_HANDLE link, stdint.uint64_t* peer_max_message_size)
     int link_set_attach_properties(LINK_HANDLE link, c_amqp_definitions.fields attach_properties)
+    int link_set_desired_capabilities(LINK_HANDLE link, c_amqpvalue.AMQP_VALUE desired_capabilities)
+    int link_get_desired_capabilities(LINK_HANDLE link, c_amqpvalue.AMQP_VALUE* desired_capabilities)
+    int link_set_offered_capabilities(LINK_HANDLE link, c_amqpvalue.AMQP_VALUE offered_capabilities)
```
annatisch: I don't think these will ever be set, only retrieved. So maybe only have the one function:
`link_get_peer_offered_capabilities`

uamqp/async_ops/client_async.py
```diff
@@ -807,6 +807,7 @@ def __init__(
                 properties=self._link_properties,
                 error_policy=self._error_policy,
                 encoding=self._encoding,
+                desired_capabilities=self._desired_capabilities,
```
annatisch: Does this need to be passed into the MessageReceiver constructor?

src/link.pyx
yunhaoling: you are right, should wrap it by calling value_factory as it's a c object.

src/link.pyx
yunhaoling: will update this

src/vendor/azure-uamqp-c/src/link.c
yunhaoling: Yes, it seems that to enable the runtime-metric feature of eventhubs, we just need to set the desired_capabilities of the link - you can see how i create the properties and set it on the receiver [here](https://github.com/Azure/azure-sdk-for-python/blob/9928d9cd31b4e880246fae8f5225c73061c8e607/sdk/eventhub/azure-eventhubs/azure/eventhub/aio/consumer_async.py#L135), this can work.
I don't use `get_desired_capabilities`.

src/vendor/inc/c_link.pxd
yunhaoling: What I am thinking is, if someone want to use our uamqp-python library to establish uamqp server and client, then we should expose both setting/getting on these two properties.

My understanding is that `offered_capabilities` is for the server role to set, and the `desired_capabilities` is for the client role to set.

uamqp/async_ops/client_async.py
yunhaoling: I'm not sure about this, I see `desired_capabilities` appears in the `ATTACH` frame described in the amqp protocol.
And I see fields like `rcv-settle-mode` are being set in the constructor, so I put it here..

I'm neural on this and okay to remove it from the constructor.
What's your thought?

src/vendor/azure-uamqp-c/src/link.c
yunhaoling: replaced indent with space, fixed

src/vendor/azure-uamqp-c/src/link.c
yunhaoling: replaced indent with space, fixed

src/vendor/azure-uamqp-c/src/link.c
yunhaoling: fixed

src/vendor/azure-uamqp-c/src/link.c
yunhaoling: fixed

src/vendor/azure-uamqp-c/src/link.c
yunhaoling: fixed

src/vendor/azure-uamqp-c/src/link.c
yunhaoling: fixed

src/vendor/azure-uamqp-c/src/link.c
yunhaoling: fixed

src/vendor/azure-uamqp-c/src/link.c
annatisch: If we don't need to get offered_capabilities - why add the extra functions to support it?

src/vendor/azure-uamqp-c/src/link.c
```diff
@@ -1083,6 +1099,84 @@ int link_get_peer_max_message_size(LINK_HANDLE link, uint64_t* peer_max_message_
...
+    }
+    else if ((link->link_state != LINK_STATE_ATTACHED) &&
+    (link->link_state != LINK_STATE_HALF_ATTACHED_ATTACH_RECEIVED))
+    {
+                LogError("Attempting to read offered capabilities before it was received");
+        result = __FAILURE__;
+    }
+    else
+    {
+        *offered_capabilities = amqpvalue_clone(link->offered_capabilities);
```
annatisch: Not quite :)
it needs to be something like this:
```
AMQP_VALUE link_capabilities = amqpvalue_clone(link->offered_capabilities);
if (link_capabilities == NULL)
{
    LogError("Failed to clone link offered capabilities");
    result = __FAILURE__;
}
else
{
    *offered_capabilities = link_capabilities ;
}
```

src/vendor/azure-uamqp-c/src/link.c
```diff
@@ -1083,6 +1099,84 @@ int link_get_peer_max_message_size(LINK_HANDLE link, uint64_t* peer_max_message_
...
+        (offered_capabilities == NULL))
+    {
+        LogError("Bad arguments: link = %p, offered_capabilities = %p",
+            link, offered_capabilities);
+        result = __FAILURE__;
+    }
+    else if ((link->link_state != LINK_STATE_ATTACHED) &&
+    (link->link_state != LINK_STATE_HALF_ATTACHED_ATTACH_RECEIVED))
+    {
+                LogError("Attempting to read offered capabilities before it was received");
```
annatisch: Some more indentation weirdness going on here....

uamqp/async_ops/client_async.py
annatisch: Nevermind - I misread the diff - this is fine I think

src/vendor/azure-uamqp-c/src/link.c
yunhaoling: I added them because the `offered_capabilities` is part of the protocol and is being missed in our implementation. I want to implement the protocol.😃

one thing I'm not doing good enough is that I didn't extract the offer_capabilities from the attach frame which makes the `desired_properties` seems useless...

maybe you're right.. I should skip it for now

yunhaoling: you're right, previously I thought this attribute would be useful, but right now it seems not.
I'll remove it.

samples/test_azure_event_hubs_receive.py
```diff
@@ -98,11 +97,12 @@ def test_event_hubs_client_proxy_settings(live_eventhub_config):
         live_eventhub_config['consumer_group'],
         live_eventhub_config['partition'])
 
-    #if not sys.platform.startswith('darwin'):  # Not sure why this passes for OSX:
+    # if not sys.platform.startswith('darwin'):  # Not sure why this passes for OSX:
     #    with pytest.raises(errors.AMQPConnectionError):
```
annatisch: Can we remove these two lines?
This passes on all platforms now?

samples/test_azure_event_hubs_receive.py
```diff
@@ -128,6 +128,40 @@ def test_event_hubs_client_receive_sync(live_eventhub_config):
...
+        log.info("Got batch: {}".format(len(batch)))
+        assert len(batch) <= 10
+        for message in batch:
+            annotations = message.annotations
+            delivery_annotations = message.delivery_annotations
+            log.info("Sequence Number: {}".format(annotations.get(b'x-opt-sequence-number')))
+            log.info("Last enqueued sequence number: {}".format(delivery_annotations.get(b'last_enqueued_sequence_number')))
+            log.info("Last enqueued offset: {}".format(delivery_annotations.get(b'last_enqueued_offset')))
+            log.info("Last enqueued time utc: {}".format(delivery_annotations.get(b'last_enqueued_time_utc')))
+            log.info("Runtime info retrieval time utc: {}".format(delivery_annotations.get(b'runtime_info_retrieval_time_utc')))
```
annatisch: These log statements should be assert statements:
`assert b'last_enqueued_sequence_number' in annotations`

samples/asynctests/test_azure_event_hubs_receive_async.py
```diff
@@ -129,6 +128,35 @@ def on_message_received(message):
...
+        message_batch = await receive_client.receive_message_batch_async(10)
+        log.info("got batch: {}".format(len(message_batch)))
+        for message in message_batch:
+            annotations = message.annotations
+            delivery_annotations = message.delivery_annotations
+            log.info("Sequence Number: {}".format(annotations.get(b'x-opt-sequence-number')))
+            log.info("Last enqueued sequence number: {}".format(delivery_annotations.get(b'last_enqueued_sequence_number')))
+            log.info("Last enqueued offset: {}".format(delivery_annotations.get(b'last_enqueued_offset')))
+            log.info("Last enqueued time utc: {}".format(delivery_annotations.get(b'last_enqueued_time_utc')))
+            log.info("Runtime info retrieval time utc: {}".format(delivery_annotations.get(b'runtime_info_retrieval_time_utc')))
```
annatisch: These log statements should be assert statements:
`assert b'last_enqueued_sequence_number' in annotations`

src/vendor/azure-uamqp-c/src/link.c
yunhaoling: fixed

src/vendor/azure-uamqp-c/src/link.c
yunhaoling: fixed

samples/test_azure_event_hubs_receive.py
yunhaoling: yes, they can all pass.
I removed these two lines.

samples/test_azure_event_hubs_receive.py
yunhaoling: test updated

samples/asynctests/test_azure_event_hubs_receive_async.py
yunhaoling: test updated

PR #96, opened 2019-09-26
Service Bus message transfer fixes

annatisch:


annatisch:
@yunhaoling - you can ignore the separate delivery tag PR and just review this one, as it has those changes as well.

KieranBrantnerMagee:
++LGTM

src/vendor/azure-uamqp-c/src/message.c
```diff
@@ -1447,3 +1465,71 @@ int message_get_message_format(MESSAGE_HANDLE message, uint32_t *message_format)
...
+    {
+        AMQP_VALUE new_delivery_tag = amqpvalue_clone(delivery_tag_value);
+        if (new_delivery_tag == NULL)
+        {
+            LogError("Cannot clone delivery tag");
+            result = __FAILURE__;
+        }
+        else
+        {
+
```
yunhaoling: should we add this here?
```c
if (message->delivery_tag != NULL)
{
    amqpvalue_destroy(message->delivery_tag);
}
```

src/vendor/azure-uamqp-c/src/message.c
```diff
@@ -1447,3 +1465,71 @@ int message_get_message_format(MESSAGE_HANDLE message, uint32_t *message_format)
...
+    else
+    {
+        if (message->delivery_tag == NULL)
+        {
+            *delivery_tag_value = NULL;
+            result = 0;
+        }
+        else
+        {
+            AMQP_VALUE new_delivery_tag = amqpvalue_clone(message->delivery_tag);
```
yunhaoling: we can just do it this way without declaring the `new_delivery_tag`?
```c
*delivery_tag_value = amqpvalue_clone(message->delivery_tag);
if(*delivery_tag_value == NULL)
{
    LogError("Cannot clone delivery tag");
    result = __FAILURE__;
}
else
{
    return = 0;
}
```

src/vendor/azure-uamqp-c/src/link.c
```diff
@@ -1575,6 +1570,12 @@ void link_dowork(LINK_HANDLE link)
     {
         tickcounter_ms_t current_tick;
 
+        if (link->current_link_credit == 0)
```
yunhaoling: I'm not sure about this one.
From the code I think what you proposed to do is by calling this way to get the sender/receiver start working:
```c
while(work_flag) {
    connection_dowork(conn);
    link_dowork(link);
}
```

the responsibility of `send_flow` is shipped from callback to an active way.

I'm wondeing about the case below:

1. The current link credit is 5
2. connection_dowork receives 10 Transfer frames
3. it will call the callback function `link_frame_received`
4. the link_instance->current_link_credit becomes < 0
5. link_dowork is called

src/vendor/azure-uamqp-c/src/message_receiver.c
```diff
@@ -236,35 +235,60 @@ static AMQP_VALUE on_transfer_received(void* context, TRANSFER_HANDLE transfer,
...
         {
-            AMQPVALUE_DECODER_HANDLE amqpvalue_decoder = amqpvalue_decoder_create(decode_message_value_callback, message_receiver);
-            if (amqpvalue_decoder == NULL)
+            delivery_tag received_message_tag;
+            if (transfer_get_delivery_tag(transfer, &received_message_tag) != 0)
             {
-                LogError("Cannot create AMQP value decoder");
-                set_message_receiver_state(message_receiver, MESSAGE_RECEIVER_STATE_ERROR);
+                    LogError("Could not get the delivery tag from the transfer performative");
+                    set_message_receiver_state(message_receiver, MESSAGE_RECEIVER_STATE_ERROR);
```
yunhaoling: indentation

src/vendor/azure-uamqp-c/src/message.c
KieranBrantnerMagee: This may be a silly reason for me to feel unsure about that suggestion, but causing a side effect (setting the output value to null) in the failure case seems a little bit "impure" given that we're communicating failure via the return variable, at least in terms of saving effectively a line of code.  If there's precedence in other functions that overrides this feel free to entirely ignore me.

yunhaoling: @KieranBrantnerMagee you're right
no problem for this one

src/message.pyx
```diff
@@ -233,6 +233,16 @@ cdef class cMessage(StructBase):
         if c_message.message_set_message_format(self._c_value, value) != 0:
                 self._value_error()
 
+    @property
+    def delivery_tag(self):
```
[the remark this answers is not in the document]
annatisch: No I don't think so - this tag is (I think) only set by a broker, and the current library has no ability to act as a broker/server

src/vendor/azure-uamqp-c/src/message.c
annatisch: I can't tell where you're suggesting I put it - I think the github diff is messed up....

src/vendor/azure-uamqp-c/src/message.c
```diff
@@ -1447,3 +1465,71 @@ int message_get_message_format(MESSAGE_HANDLE message, uint32_t *message_format)
...
+int message_set_delivery_tag(MESSAGE_HANDLE message, AMQP_VALUE delivery_tag_value)
+{
+    int result;
+
+    if ((message == NULL) || (delivery_tag_value == NULL))
+    {
+        LogError("Bad arguments: message = %p, delivery_tag = %p",
+            message, delivery_tag_value);
+        result = __FAILURE__;
+    }
```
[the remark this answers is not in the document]
annatisch: Interesting - I don't think this will be an issue here because we're already NULL validating the value before this function, and also this value should never be set twice on a single message (making the reset unnecessary). However I have adopted the same behaviour to be consistent.

src/vendor/azure-uamqp-c/src/message.c
annatisch: Ah I get it - again I don't think it's necessary as this value should never be set twice, but will add to be extra safe

src/vendor/azure-uamqp-c/src/link.c
annatisch: Yeah - tbh I don't know about this either.... And I need to spend more time testing it.
At the very least I can make this a `<= 0` check in the meantime, in case we ever hang because it somehow went negative...

src/message.pyx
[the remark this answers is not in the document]
yunhaoling: oh, I thought that we could act as a server, good to know that we don't have the ability right now.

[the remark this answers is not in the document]
annatisch: It's true that the C library can - however those features were never exposed in the Python bindings :)

PR #98, opened 2019-10-01
Performance improvement

yunhaoling:
Lazy parse message

### DIFF
--- a/src/amqpvalue.pyx
+++ b/src/amqpvalue.pyx
@@ -313,7 +313,7 @@ cdef class AMQPValue(StructBase):
         if <void*>value == NULL:
             self._value_error()
         as_string = c_amqpvalue.amqpvalue_to_string(value)
-        py_string = copy.deepcopy(as_string)
+        py_string = as_string
         c_amqpvalue.amqpvalue_destroy(self._c_value)
         return py_string
 
@@ -669,7 +669,7 @@ cdef class StringValue(AMQPValue):
         assert self.type
         cdef const char* _value
         if c_amqpvalue.amqpvalue_get_string(self._c_value, &_value) == 0:
-            return copy.deepcopy(_value)
+            return _value
         else:
             self._value_error()
 
@@ -687,7 +687,7 @@ cdef class SymbolValue(AMQPValue):
         assert self.type
         cdef const char* _value
         if c_amqpvalue.amqpvalue_get_symbol(self._c_value, &_value) == 0:
-            return copy.deepcopy(_value)
+            return _value
         else:
             self._value_error()
 
@@ -742,7 +742,7 @@ cdef class ListValue(AMQPValue):
         assert self.type
         value = []
         for i in range(self.size):
-            value.append(copy.deepcopy(self[i].value))
+            value.append(self[i].value)
         return value
 
 
@@ -795,7 +795,7 @@ cdef class DictValue(AMQPValue):
         value = {}
         for i in range(self.size):
             key, item = self.get(i)
-            value[copy.deepcopy(key.value)] = copy.deepcopy(item.value)
+            value[key.value] = item.value
         return value
 
 
@@ -840,7 +840,7 @@ cdef class ArrayValue(AMQPValue):
         assert self.type
         value = []
         for i in range(self.size):
-            value.append(copy.deepcopy(self[i].value))
+            value.append(self[i].value)
         return value
 
 
@@ -945,4 +945,4 @@ cdef class DescribedValue(AMQPValue):
         assert self.type
         #descriptor = self.description
         described = self.data
-        return copy.deepcopy(described.value)
+        return described.value
--- a/src/amqpvalue.pyx
+++ b/src/amqpvalue.pyx
@@ -8,6 +8,7 @@
 from enum import Enum
 import logging
 import uuid
+import copy
 
 import six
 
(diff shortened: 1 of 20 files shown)


src/amqpvalue.pyx
```diff
@@ -313,7 +313,7 @@ cdef class AMQPValue(StructBase):
         if <void*>value == NULL:
             self._value_error()
         as_string = c_amqpvalue.amqpvalue_to_string(value)
-        py_string = copy.deepcopy(as_string)
```
annatisch: Perhaps we can remove `import copy`

src/message.pyx
```diff
@@ -7,6 +7,7 @@
 # Python imports
 from enum import Enum
 import logging
+import message
```
annatisch: This doesn't seem to be used?

uamqp/message.py
```diff
@@ -131,12 +192,41 @@ def __str__(self):
             return ""
         return str(self._body)
 
-    def _parse_message(self, message):
+    def _parse_message_properties(self):
```
annatisch: We could probably improve things further be breaking this down to only load the specific properties on demand (e.g. only delivery annotations) - however I think this is a good approach for an incremental change given then time constraints.

yunhaoling:
@annatisch Please combine this with the PR in eventhub:https://github.com/Azure/azure-sdk-for-python/pull/7533 to see the whole picture.

src/message.pyx
yunhaoling: removed

uamqp/message.py
yunhaoling: yes, I agree. when I implement this, i thought we could be doing parse for each property without parsing them all.

src/amqpvalue.pyx
yunhaoling: right, removed

uamqp/message.py
YijunXieMS: Aside from lazy loading, let's try to tune the parsing code itself in the next preview.

uamqp/message.py
```diff
@@ -109,10 +110,81 @@ def __init__(self,
...
+    def properties(self):
+        if self._need_further_parse:
+            self._parse_message_properties()
+        return self._properties
+
+    @properties.setter
+    def properties(self, value):
+        if not isinstance(value, MessageProperties):
+            raise TypeError("Properties must be a MessageProperties.")
+        self._message.properties = value.get_properties_obj()
```
annatisch: We don't need to do this here because it's done in `def _populate_message_attributes` which is called during send/encode

uamqp/message.py
```diff
@@ -109,10 +110,81 @@ def __init__(self,
...
+    def header(self):
+        if self._need_further_parse:
+            self._parse_message_properties()
+        return self._header
+
+    @header.setter
+    def header(self, value):
+        if not isinstance(value, MessageHeader):
+            raise TypeError("Header must be a MessageHeader.")
+        self._message.header = value.get_header_obj()
```
annatisch: We don't need to do this here because it's done in `def _populate_message_attributes` which is called during send/encode

uamqp/message.py
```diff
@@ -109,10 +110,81 @@ def __init__(self,
...
+        if self._need_further_parse:
+            self._parse_message_properties()
+        return self._application_properties
+
+    @application_properties.setter
+    def application_properties(self, value):
+        if not isinstance(value, dict):
+            raise TypeError("Application properties must be a dictionary.")
+        amqp_props = utils.data_factory(value, encoding=self._encoding)
+        self._message.application_properties = amqp_props
```
annatisch: We don't need to do this here because it's done in `def _populate_message_attributes` which is called during send/encode

uamqp/message.py
```diff
@@ -109,10 +110,81 @@ def __init__(self,
...
+            self._parse_message_properties()
+        return self._annotations
+
+    @annotations.setter
+    def annotations(self, value):
+        if not isinstance(value, dict):
+            raise TypeError("Message annotations must be a dictionary.")
+        ann_props = c_uamqp.create_message_annotations(
+            utils.data_factory(value, encoding=self._encoding))
+        self._message.message_annotations = ann_props
```
annatisch: We don't need to do this here because it's done in `def _populate_message_attributes` which is called during send/encode

uamqp/message.py
annatisch: It occurs to me that doesn't the EH consumer _always_ use the message properties to get the offset while it's running?
In which case isn't the lazy loading somewhat ignored?
Or does the consumer not pull the annotations for _every_ message?

yunhaoling: aha, you got the point here!

Previously we update consumer.offset every time we create an event from a message. But I find that we don't need to update the consumer.offset such frequently. We just need to update the offset according to the last offset.

So when we receive 100 msgs, we just need the offset information of the last msg, not each one.

https://github.com/Azure/azure-sdk-for-python/blob/647b2561318ae127870b0ca503b09b5fbe09a7f9/sdk/eventhub/azure-eventhubs/azure/eventhub/aio/consumer_async.py#L203-L209

uamqp/message.py
yunhaoling: thanks for point out this, removed

uamqp/message.py
yunhaoling: thanks for point out this, removed

uamqp/message.py
yunhaoling: thanks for point out this, removed

uamqp/message.py
yunhaoling: thanks for point out this, removed

PR #106, opened 2019-10-14
Multiple session sharing connection

yunhaoling:
- [X] C implementation
- [x] C Test case
- [x] Python implementation
- [x] Python Test/Sample case


uamqp/async_ops/connection_async.py
```diff
@@ -98,6 +99,15 @@ def __init__(self, hostname, sasl,
         await self.destroy_async()
         _logger.debug("Finished exiting connection %r context.", self.container_id)
 
+    async def _open_async(self):
```
KieranBrantnerMagee: Should we allow for a timeout to be passed? (re: potentially looping forever)

src/vendor/azure-uamqp-c/samples/multiple_receivers_share_connection_sample/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+
+compileTargetAsC99(multiple_receivers_share_connection_sample)
+
+include_directories(.)
+if (WIN32)
+	#windows needs this define
+	add_definitions(-D_CRT_SECURE_NO_WARNINGS)
+	add_definitions(-DGB_MEASURE_MEMORY_FOR_THIS -DGB_DEBUG_ALLOC)
+
+if(WINCE)
```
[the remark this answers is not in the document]
yunhaoling: Fixed, The indentation in VS/VSC is by default TAB.. changed to 4 spaces

src/vendor/azure-uamqp-c/samples/multiple_receivers_share_connection_sample/main.c
```diff
@@ -0,0 +1,255 @@
...
+
+    MESSAGE_RECEIVER_HANDLE* receiver_handlers = malloc(sizeof(MESSAGE_RECEIVER_HANDLE) * partition_cnt);
+    SESSION_HANDLE* session_handlers = malloc(sizeof(SESSION_HANDLE) * partition_cnt);
+    LINK_HANDLE* link_handlers = malloc(sizeof(LINK_HANDLE) * partition_cnt);
+
+    int* running_context = malloc(sizeof(int) * partition_cnt);
+
+    for (int i = 0; i < partition_cnt; i++) {
+	    session_handlers[i] = create_session(connection);
+        LINK_HANDLE* created_link = NULL;
```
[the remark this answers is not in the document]
yunhaoling: I add more comments to help understand the code.
This c code is basically showing how to multiplex uamqp connection and use that for receiving, not targeting at showing EH functionality - I copied most part of the code from other c sample.😀

However, it can still be improved as you said, in the future time when we have more time

uamqp/async_ops/connection_async.py
yunhaoling: you're right, I need to check whether the underlying c/cython code would take care of the open timeout. If not then I need to add check here.

samples/asynctests/test_azure_iothub_cli_extension.py
```diff
@@ -335,6 +335,7 @@ def get_target(config):
     return target
 
 def test_iothub_monitor_events(live_iothub_config):
+    pytest.skip('Unable to open connection/socket with the given hostname, skipping for now')
```
yunhaoling: @annatisch , This test case would raise `RuntimeError: Operation failed` in my current implementation.
So I tested this livetest over uamqp1.2.0 and 1.2.3. I found that this test case actually is always failing, but exceptions are just not raised.

I would prefer skipping it for now.

Here is the log information:
[error information on current implementation.txt](https://github.com/Azure/azure-uamqp-python/files/3737740/error.information.on.current.implementation.txt)
[error information on uamqp1.2.0 and 1.2.3.txt](https://github.com/Azure/azure-uamqp-python/files/3737741/error.information.on.uamqp1.2.0.and.1.2.3.txt)

src/vendor/azure-uamqp-c/src/session.c
```diff
@@ -862,15 +862,30 @@ int session_begin(SESSION_HANDLE session)
...
-                    result = __FAILURE__;
+                    session_instance->is_underlying_connection_open = UNDERLYING_CONNECTION_OPEN;
+                    result = 0;
                 }
-                else
+                else if(connection_open_result == CONNECTION_OPEN_ALREADY_OPEN)
                 {
                     session_instance->is_underlying_connection_open = UNDERLYING_CONNECTION_OPEN;
-                    result = 0;
+                    if (send_begin(session) != 0)
```
annatisch: Why do we only call `send_begin` if `connection_open_result == CONNECTION_OPEN_ALREADY_OPEN` and not if `connection_open_result == CONNECTION_OPEN_OK`?

uamqp/async_ops/client_async.py
```diff
@@ -276,7 +279,7 @@ def __init__(
             self._keep_alive_thread = None
         if not self._session:
             return  # already closed.
-        if not self._connection.cbs:
+        if not self._using_cbs_session:
```
annatisch: Even if we're not use the CBS session for sending/receiving - do we really want to destroy it? Isn't the CBS session used to authenticate all the other sessions using the same connection?

uamqp/async_ops/connection_async.py
```diff
@@ -98,6 +99,15 @@ def __init__(self, hostname, sasl,
...
+        self._conn.open()
+        connection_state = c_uamqp.ConnectionState(self._conn.get_state())
+        while connection_state != c_uamqp.ConnectionState.OPENED:
```
annatisch: I feel like this could be made more resiliant... perhaps something like:
```
while connection_state not in [OPENED, ERROR, etc etc]
# or the reverse
while connection_state in [list of all the opening states]
```

uamqp/async_ops/client_async.py
annatisch: Same comment for sync client.

uamqp/async_ops/connection_async.py
annatisch: Same comment for sync connection

src/vendor/azure-uamqp-c/src/session.c
yunhaoling: The `send_begin` function can only be called when the connection has been already opened, that is the connection state being `CONNECTION_STATE_OPENED`.

If it's a new session on new connection, the `send_begin` will be called in the [`on_connection_state_changed`](https://github.com/Azure/azure-uamqp-python/blob/0a631b33f46ccbea25ffa15d811e5eb3a4a7edc4/src/vendor/azure-uamqp-c/src/session.c#L363).

However, if it's an existing opened connection, there is no chance of running into the `on_connection_state_changed` and newew session couldn't be begun - that's the reason we previously don't support multiple sessions on one connection and why we only need the `session_begin` for newer sessions in `CONNECTION_OPEN_ALREADY_OPEN` case.

But overall the `connection_open_result` is just an indicator for the creation result of the connection object. It still can't indicate the connection_state, which means the `send_begin` could fail if a connection has been created but too hurry to finish the processing of opening(shaking hands with the server side).

There is certain situation that multiple sessions are started on one single connection simultaneously, the relatively-latter session would check and find connection already created while the underlying connection has not finished shaking hands. So this would lead to session begin failure.

To handle the situation I described above, I think it's the application-level responsibility to ensure the connection is opened before we want to multiplexing it. So I put extra logic in the context manager and `redirect` of connection to get it opened first.

Well, this is the whole story of enabling multiple sessions. Hope it won't confuse you a lot.

Let me know if you have more questions.

uamqp/async_ops/client_async.py
yunhaoling: I think this can be spitted into two questions:

Q1: Do we need to destroy these normal sessions?

The answer is: Yes.

1. If we don't destroy these normal sessions here, then there is no other place to handle these useless sessions (to free the underlying c objects). Unless we keep tracking sessions on a connection, and release all the sessions when we're shutting down a connection.

2. Consider the situation we're using SASLPlain, there would be no cbs session in this case, just normal sessions.

Q2: How the CBS session authenticates other sessions?

The answer may surprise you, when I was playing with C code, I found in eventhub use case we don't need to use cbs session to authenticate for other sessions. We just need to put the token one time at the very beginning then every following session can work - no need to put extra tokens for those new session.

Personally I think it's the generated token authenticates access to all the resources under certain eventhub namespace.

uamqp/async_ops/connection_async.py
yunhaoling: Code updated.
also added new parameter `open_timeout` in the connection constructor and `timeout` in `open` function

as @KieranBrantnerMagee mentioned his concern on the looping forever in the `open` function. I checked the underlying `socketio_dowork` but find nothing related to timeout control.

uamqp/async_ops/connection_async.py
yunhaoling: I checked the underlying socketio_dowork but find nothing related to timeout control.
added timeout control here.

samples/asynctests/test_azure_iothub_cli_extension.py
digimaun: The Az CLI IoT extension uses uamqp for all amqp operations. 

Can we please make sure python uamqp is compatible with the CLI IoT extension use cases and if there are some breaking changes or design change that needs to be updated in the IoT extension you can let me know? In the past @annatisch maintained this partnership (as this test shows). It seems suspicious that with this update the test is skipped.

Interesting datapoint from @yunhaoling that this test is always failing in 1.2.0 and 1.2.3. We are currently using v1.1.0 - we tried updating to 1.2.4  to support Py 3.8 but ran into issues, such as segfaults in Mac environment. We planned to try with 1.2.5 or a later version in the coming days/weeks. 

What changed from 1.1 -> 1.2 that are causing these exceptions?

PR #112, opened 2019-11-15
Uamqp updates

yunhaoling:
- [x] Remove executor
- [x] Call private encode message
- [x] Timeout calculation in milliseconds

### DIFF
--- a/uamqp/message.py
+++ b/uamqp/message.py
@@ -557,7 +557,10 @@ class BatchMessage(Message):
                     try:
                         if not data.application_properties and self.application_properties:  # Message-like object
                             data.application_properties = self.application_properties
-                        message_bytes = data._encode_message()  # pylint: disable=protected-access
+                        try:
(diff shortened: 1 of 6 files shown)


uamqp/message.py
```diff
@@ -596,7 +596,7 @@ def gather(self):
             try:
                 if not data.application_properties and self.application_properties:  # Message-like object
                     data.application_properties = self.application_properties
-                message_bytes = data.encode_message()
+                message_bytes = data._encode_message()  # pylint: disable=protected-access
```
annatisch: We need to support both `encode_message` and `_encode_message`

yunhaoling: you're right, otherwise we lose compatibility.

annatisch:
Do you want to add the version bump/release notes etc to this PR or separately?

yunhaoling:
@annatisch  we can do this separately

PR #118, opened 2019-11-21
batch application_properties implementation update

yunhaoling:


uamqp/message.py
```diff
@@ -555,12 +555,10 @@ def _multi_message_generator(self):
                 for data in self._body_gen:
                     message_bytes = None
                     try:
-                        if not data.application_properties and self.application_properties:  # Message-like object
-                            data.application_properties = self.application_properties
-                        try:
-                            message_bytes = data.encode_message()
-                        except AttributeError:
-                            message_bytes = data._encode_message()  # pylint: disable=protected-access
+                        internal_uamqp_message = data.message  # Data object has an internal uamqp message
```
annatisch: I think I would really prefer:
```
try:
    internal_uamqp_message = data.message
except AttributeError:
    internal_uamqp_message = data
```
Just to be extra safe....

yunhaoling: updated the code, it's stronger than before to handle all the case.

PR #121, opened 2019-11-22
Update windows env

bsiegel:
To build with VS2019, we need setuptools >=34.4.0. Also in order to select the target platform we need to use the `-A` flag now rather than simply appending `Win64` to the Generator name. This means we now have an implicit requirement on CMake 3.1 (because that's when the `-A` flag was added).

Exploratory PR for #122

### DIFF
--- a/.azure-pipelines/client.openssl.yml
+++ b/.azure-pipelines/client.openssl.yml
@@ -2,13 +2,13 @@ trigger: none
 pr: none
 
 variables:
-  OpenSSLVersion: '1.0.2q'
+  OpenSSLVersion: '1.0.2t'
 
 jobs:
   - job: 'OpenSSL'
 
     pool:
-      vmImage: 'macOS-10.13'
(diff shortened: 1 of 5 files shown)


bsiegel:
This -should- be enough to get things working, but it's not. The issues remaining seem to be:

Python 3.5:
It's unable to locate a header file. My initial scan of this seems to indicate that as compared to the other Python version builds, the call to `cl.exe` is missing a number of include directories. This leads me to think that something in this version of Python (maybe distutils?) is too old to know how to locate those directories. We might have to figure out how to add them manually or whether there is just some component we can update with pip.

Python 3.6 x86:
This is odd because the x64 build passes. The x86 build always fails with a linker error, `unresolved external symbol __except_handler4_common`. I've tried using the `-T host=x86` CMAKE flags to use the 32-bit compiler toolchain, the flag worked as documented but that didn't resolve the issue. At this point I'm a little stumped as to why this is happening and only for the x86 build.

PR #131, opened 2019-12-07
Connection idle timeout patch

yunhaoling:
The PR is to handle situation when connection state changed from XXX -> DISCARDING.

When connection idle timeout happens, the underlying C code would set the connection status to DISCARDING.

In previous implementation, the connection status would go through weird transition in the case of idle timeout:
```
OPENED->DISCARDING
DISCARDING->DISCARDING
DISCARDING->DISCARDING
...
DISCARDING->DISCARDING
DISCARDING->END
END->DISCARDING # This is the most weird part
```

Although the functionality is not affected, but the improper status transition is annoying.
This can be improved in `state_changed` of connection.py to handle the case of DISCARDING.


annatisch:
@yunhaoling - this patch looks good to me - is there any way we can add a test for it?

PR #132, opened 2019-12-09
Updates for 1.2.5

yunhaoling:
Mostly it's the same PR as https://github.com/Azure/azure-uamqp-python/pull/127.

Created for livetest/documentation update/any other missing changes.

### DIFF
--- a/samples/test_azure_event_hubs_receive.py
+++ b/samples/test_azure_event_hubs_receive.py
@@ -37,6 +37,18 @@ def get_plain_auth(config):
         config['access_key'])
 
 
+def send_single_message(live_eventhub_config, partition, msg_content):
+    target = "amqps://{}/{}/Partitions/{}".format(
+        live_eventhub_config['hostname'],
+        live_eventhub_config['event_hub'],
+        partition
+    )
+    uri = "sb://{}/{}".format(live_eventhub_config['hostname'], live_eventhub_config['event_hub'])
+    sas_auth = authentication.SASTokenAuth.from_shared_access_key(
+        uri, live_eventhub_config['key_name'], live_eventhub_config['access_key'])
+    uamqp.send_message(target, msg_content, auth=sas_auth, debug=False)
+
+
 def test_event_hubs_simple_receive(live_eventhub_config):
     source = "amqps://{}/{}/ConsumerGroups/{}/Partitions/{}".format(
         live_eventhub_config['hostname'],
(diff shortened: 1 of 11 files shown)


samples/test_azure_event_hubs_receive.py
```diff
@@ -236,6 +236,32 @@ def test_event_hubs_iter_receive_sync(live_eventhub_config):
...
+        live_eventhub_config['consumer_group'])
+
+    with uamqp.Connection(live_eventhub_config['hostname'], sas_auth, debug=False) as conn:
+        partition_0 = uamqp.ReceiveClient(source + "0", debug=False, auth=sas_auth, timeout=3000, prefetch=10)
+        partition_1 = uamqp.ReceiveClient(source + "1", debug=False, auth=sas_auth, timeout=3000, prefetch=10)
+        partition_0.open(connection=conn)
+        partition_1.open(connection=conn)
+
+        try:
+            messages_0 = partition_0.receive_message_batch(1)
```
KieranBrantnerMagee: From my comment in the other pr, should send before receiving to not just assume messages were sent from other tests.

uamqp/async_ops/client_async.py
```diff
@@ -313,16 +314,18 @@ def __init__(
...
-            operation,
-            op_type=op_type,
-            node=node,
-            callback=callback,
-            encoding=self._encoding,
-            debug=self._debug_trace,
-            **kwargs))
+            await asyncio.sleep(0.05, loop=self.loop)
+        response = await asyncio.shield(
+            self._session.mgmt_request_async(
```
KieranBrantnerMagee: do you need to pass the loop in here?

uamqp/async_ops/client_async.py
```diff
@@ -517,7 +520,9 @@ def __init__(
         return True
 
     async def _transfer_message_async(self, message, timeout):
-        sent = await asyncio.shield(self.message_handler.send_async(message, self._on_message_sent, timeout=timeout))
+        sent = await asyncio.shield(
+            self.message_handler.send_async(message, self._on_message_sent, timeout=timeout),
```
KieranBrantnerMagee: loop threadthrough?

uamqp/async_ops/connection_async.py
```diff
@@ -98,6 +98,9 @@ def __init__(self, hostname, sasl,
     async def _close_async(self):
         _logger.info("Shutting down connection %r.", self.container_id)
         self._closing = True
+        if self._cbs:
+            await self.auth.close_authenticator_async()
```
KieranBrantnerMagee: loop?

uamqp/async_ops/client_async.py
yunhaoling: pass `loop` into the `self._session.mgmt_request_async`? That method don't need a loop parameter to be passed in.
it would be using `self.loop` in the Session

uamqp/async_ops/client_async.py
yunhaoling: loop is not used in `send_async`.
if you mean `asyncio.shield` it has already been added.

uamqp/async_ops/connection_async.py
yunhaoling: loop is not used in `close_authenticator_async`

samples/test_azure_event_hubs_receive.py
yunhaoling: Fixed, added sending

PR #133, opened 2019-12-10
Bug fix

yunhaoling:
1. renamed _build_session to _build_session_async in AsyncClient
2. cbs session initialized twice when connection is shared

### DIFF
--- a/samples/test_azure_event_hubs_receive.py
+++ b/samples/test_azure_event_hubs_receive.py
@@ -236,6 +236,32 @@ def test_event_hubs_iter_receive_sync(live_eventhub_config):
     receive_client.close()
 
 
+def test_event_hubs_shared_connection(live_eventhub_config):
+    uri = "sb://{}/{}".format(live_eventhub_config['hostname'], live_eventhub_config['event_hub'])
+    sas_auth = authentication.SASTokenAuth.from_shared_access_key(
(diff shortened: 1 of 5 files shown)


samples/test_azure_event_hubs_receive.py
```diff
@@ -236,6 +236,32 @@ def test_event_hubs_iter_receive_sync(live_eventhub_config):
...
+        live_eventhub_config['consumer_group'])
+
+    with uamqp.Connection(live_eventhub_config['hostname'], sas_auth, debug=False) as conn:
+        partition_0 = uamqp.ReceiveClient(source + "0", debug=False, auth=sas_auth, timeout=3000, prefetch=10)
+        partition_1 = uamqp.ReceiveClient(source + "1", debug=False, auth=sas_auth, timeout=3000, prefetch=10)
+        partition_0.open(connection=conn)
+        partition_1.open(connection=conn)
+
+        try:
+            messages_0 = partition_0.receive_message_batch(1)
```
KieranBrantnerMagee: probably best to not assume the EH has events prior, in case this is run first for whatever reason.

yunhaoling: seems so, then we need to update other tests as well.
I'd like to have tests improved in the next release.

PR #142, opened 2020-01-17
Support for connection desired capabilities

yunhaoling:
The desired capabilities of a connection is described in the amqp protocol, the information is put into the `OPEN` frame.

However, it's not able to verify this feature -- currently the iothub redirect takes place even `amqp:link:redirect` is not put into the desired-capabilities. But the service side requires this as in the future they would check the information.

for more context:
redirect could happen on link or connection level
could be related to EH redirect support

### DIFF
--- a/src/connection.pyx
+++ b/src/connection.pyx
@@ -171,6 +171,21 @@ cdef class Connection(StructBase):
             self._value_error()
 
     @property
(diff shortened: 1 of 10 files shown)


annatisch:
This patch looks good too - can we add a test using the amqp:link:redirect mentioned above?

PR #143, opened 2020-02-06
Update tlsio options when creating tlsio on macos

yunhaoling:
Related Issue: https://github.com/Azure/azure-sdk-for-python/issues/7201

### DIFF
--- a/src/vendor/azure-uamqp-c/deps/azure-c-shared-utility/pal/ios-osx/tlsio_appleios.c
+++ b/src/vendor/azure-uamqp-c/deps/azure-c-shared-utility/pal/ios-osx/tlsio_appleios.c
@@ -258,7 +258,7 @@ static CONCRETE_IO_HANDLE tlsio_appleios_create(void* io_create_parameters)
                     result->sockWrite = NULL;
                     result->hostname = NULL;
                     result->pending_transmission_list = NULL;
-                    tlsio_options_initialize(&result->options, TLSIO_OPTION_BIT_NONE);
+                    tlsio_options_initialize(&result->options, TLSIO_OPTION_BIT_TRUSTED_CERTS);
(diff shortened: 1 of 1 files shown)


bsiegel:
We should work to upstream this change back to the original repo: https://github.com/Azure/azure-c-shared-utility/

Then of course the consuming library https://github.com/Azure/azure-uamqp-c will need to update its submodules, and then finally we will need to update our vendored copy of azure-uamqp-c

bsiegel:
Probably want to wait on merging this and actually update our vendored copy of azure-uamqp-c, once it has updated *its* submoduled copy of azure-c-shared-utility

yunhaoling:
@bsiegel agreed, thanks for the suggestion.

I've already created a PR in the azure-c-shared-utility:
https://github.com/Azure/azure-c-shared-utility/pull/427

but yet no one is reviewing it, I'll ping someone to see if I can get my PR reviewed.

malthe:
@bsiegel this seems to have been merged upstream now.

bsiegel:
Upstream azure-uampq-c library needs to pull in the updated submodule (issue is open, https://github.com/Azure/azure-uamqp-c/issues/331). Once that's done we can then update our copy of azure-uamqp-c and unblock this.

yunhaoling:
hey @bsiegel, actually we have deviated from azure-uamqp-c library -- we've made our own fixes in the C lib previously and not pulled their latest code in.

Probably in the future we could make a big PR to their repo including all our fixes. But right now I prefer our release plan not being blocked by the underlying C library.

bsiegel:
Sounds good. I don't have a strong opinion since I don't actually own any code in this repo, but we should probably document somewhere the fixes we have applied on top of our dependencies, and which of those fixes have been accepted upstream. And also should file tasks to update those dependencies to the newer versions and re-apply any fixes that were not accepted upstream.

PR #157, opened 2020-04-28
Cbs auth and management support

yunhaoling:


uamqp/session.py
```diff
@@ -239,7 +243,8 @@ def _outgoing_disposition(self, frame):
 
     def _incoming_disposition(self, frame):
         for link in self._input_handles.values():
-            link._incoming_disposition(frame)
+            if hasattr(link, '_incoming_disposition'):
```
annatisch, line 246: Rather than do this check - we should just add an empty `_incoming_disposition` function to the base link class that just does nothing. I think that might be more efficient than the hasattr check

uamqp/session.py
```diff
@@ -321,3 +326,30 @@ def create_request_response_link_pair(self, endpoint, **kwargs):
             network_trace=kwargs.pop('network_trace', self.network_trace),
             **kwargs)
 
+    def mgmt_request(self, message, operation=None, operation_type=None, node='$management', **kwargs):
```
annatisch, line 329: I would rather not have this function.
I think Management link opening should be handled the same way link opening in handled, then requests should be passed straight  to the management link in the same was that outgoing messages are passed to the link.
We could put logic like this into the clients though. As they are a higher-level implementation.

uamqp/management_link.py
```diff
@@ -4,92 +4,185 @@
...
     """
 
     """
-    def __init__(self, session, endpoint, **kwargs):
+    def __init__(
+            self,
+            session,
+            endpoint,
+            status_code_field=b'statusCode',
+            status_description_field=b'statusDescription',
```
annatisch, line 36: These two should be moved into kwargs.
They only exist to account for a bug in the AMQP spec/service implementation

uamqp/management_link.py
```diff
@@ -4,92 +4,185 @@
...
+                self.state = ManagementLinkState.IDLE
+                self._on_amqp_management_open_complete(ManagementOpenResult.ERROR)
+        elif self.state == ManagementLinkState.OPEN:
+            if new_state is not LinkState.ATTACHED:
+                self.state = ManagementLinkState.ERROR
+                self._on_amqp_management_error()
         elif self.state == ManagementLinkState.CLOSING:
-            if new_state not in [LinkState.DETACHED, LinkState.DETACH_RCVD]:
-                self._set_state(ManagementLinkState.ERROR)
+            if new_state not in [LinkState.DETACHED, LinkState.DETACH_SENT, LinkState.DETACH_RCVD]:
```
annatisch, line 90: DETACH_SENT/DETACH_RCVD don't actually work yet..... The C implementation only had one state to represent these two - which is how I started - but then though I might introduce this as two separate states. No need to change this now - just a headsup

PR #158, opened 2020-04-29
Updates for v1.2.7

yunhaoling:


yunhaoling:
previous travis-ci check might have been broken due to the floated pytest and asyncio deps (probably a bad release of pytest / pytest asyncio)

example failure here: https://travis-ci.org/github/Azure/azure-uamqp-python/jobs/681173654

as per discussion with @scbedd:
update the pytest/pytest asyncio dependency (consistent with the version we currently use in our devops ci).

Question: @annatisch would this have a side effect if we're specifying "pytest == [version]" not "pytest >= [version]" which enforced our users to use a specific pytest version. (This would only affect the people who want to do development in uamqp library)

annatisch:
I don't think there should be any concerns with updating the pytest dependency.
Those doing development should be doing so within a virtual environment, rather than installing pytest into a global environment - so what we set it to should have an impact on other projects they may use pytest for.

PR #159, opened 2020-05-06
Initialize delivery_count header at 0 instead of None.

KieranBrantnerMagee:
Add changelog note detailing this.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,7 +3,7 @@
 Release History
 ===============
 
-1.2.8 (2020-05-18)
+1.2.8 (unreleased)
 ++++++++++++++++++
 - Fix to initialize delivery_count header at 0 instead of None (azure-sdk-for-python issue #9708)
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,13 +3,18 @@
 Release History
 ===============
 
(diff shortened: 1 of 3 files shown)


annatisch:
I'm not convinced this fix should be in uamqp. By setting it to 0 here, we imply that the service actually gave us a value, when in fact it was omitted.
I would prefer this fix to be in the SB SDK, where we know the exact intentions of the service concerning delivery count.

annatisch:
Ignore my earlier comment. AMQP protocol specification determines that 0 is a default value for the delivery count.

yunhaoling:
as per discussion OOB, tweak the test to accommodate encoding size change brought by delivery_count default value "None" -> "0"

HISTORY.rst
```diff
@@ -3,13 +3,18 @@
 Release History
 ===============
 
+1.2.8 (2020-05-18)
```
yunhaoling, line 6: 1.2.8 (unreleased)

yunhaoling, line 6:
```suggestion
1.2.8 (unreleased)
```

yunhaoling: (we can update the release date before we decide to release to not constrain our plan 😀)

PR #160, opened 2020-05-07
Dead Letter Fix

yunhaoling:


src/vendor/azure-uamqp-c/src/messaging.c
```diff
@@ -188,11 +188,20 @@ AMQP_VALUE messaging_delivery_rejected(const char* error_condition, const char*
...
+                    {
+                        if (rejected_set_error(rejected, error_handle) != 0)
+                        {
+                            LogError("Cannot set error on REJECTED state handle");
+                            error_constructing = true;
+                        }
+                    }
                 }
 
                 error_destroy(error_handle);
```
annatisch, line 207: Do we need to destroy the fields as well?
We should confirm whether `error_set_info` does an `amqpvalue_clone` - and if it does we will need to destroy this copy of it I think.

tests/test_c_messaging.py
```diff
@@ -49,13 +49,22 @@ def test_delivery_accepted():
 
 
 def test_delivery_rejected():
-    rej_val = c_uamqp.Messaging.delivery_rejected(b'Failed', b'Test failure')
```
annatisch, line 52: We should also keep the original test - these extra fields should be optional, and default to None.

uamqp/errors.py
```diff
@@ -267,7 +266,7 @@ class MessageAccepted(MessageResponse):
 
 class MessageRejected(MessageResponse):
 
-    def __init__(self, condition=None, description=None, encoding='UTF-8'):
+    def __init__(self, condition=None, description=None, info=None, encoding='UTF-8'):
```
annatisch, line 269: Currently this would be breaking, if anyone had specified the call like:
`MessageRejected("foo", "bar", "ascii")`

And it seems that it's technically part of the public API.... however I think we're the only ones who call this right?
Users I think would only ever call `Message.reject()`?

yunhaoling: (I tried to make this feature non-breaking.. but looks like I still miss this here😣)

AFAIK, iot-sdk is depending on our uamqp library, chances are that they might use this API?
There're options if our intention is just for non-breaking:
- move to **kwargs
- putting `info` after `encoding`

src/vendor/azure-uamqp-c/src/messaging.c
yunhaoling: I don't think we need to destroy `fields` here, `error_set_info` clean up the resources it creates.
https://github.com/Azure/azure-uamqp-python/blob/bd30f8f3b6a8f5708858409367ef335ec8ffce1e/src/vendor/azure-uamqp-c/src/amqp_definitions.c#L571-L603

and it's an object being passed from the python layer, so python would garbage collect that.

(And I checked `AMQP_VALUE messaging_delivery_modified(bool delivery_failed, bool undeliverable_here, fields message_annotations)`, it doesn't destroy message_annotations)

tests/test_c_messaging.py
yunhaoling: added back, thanks for reminding

uamqp/errors.py
annatisch: I'm pretty sure this is fine actually - I don't think a user would ever call this - and in the case of EH/IoT - I don't believe they ever reject messages anyway as they're already settled.

Even if a user did construct this object directly - I don't think there's any public API that accepts it as input....

yunhaoling: if that's the case, then I'm good with keeping my changes.

As the only internal place to construct `errors.MessageRejected` in uamqp is in `message.py:Message.reject` which makes me feel less concerned about this change.

src/message_receiver.pyx
```diff
@@ -84,9 +84,14 @@ cdef class cMessageReceiver(StructBase):
             raise RuntimeError("Unable to send message dispostition 'released' for message number {}".format(message_number))
         c_amqpvalue.amqpvalue_destroy(delivery_state)
 
-    cpdef settle_rejected_message(self, c_amqp_definitions.delivery_number message_number, const char* error_condition, const char* error_description):
+    cpdef settle_rejected_message(self, c_amqp_definitions.delivery_number message_number, const char* error_condition, const char* error_description, AMQPValue error_info=None):
         cdef c_amqpvalue.AMQP_VALUE delivery_state
-        delivery_state = c_message.messaging_delivery_rejected(error_condition, error_description)
+        cdef c_amqp_definitions.fields delivery_fields
+        if error_info is not None:
+            delivery_fields = <c_amqp_definitions.fields>error_info._c_value
```
annatisch, line 91: It looks like we have two separate code paths for rejecting a message?
Is one of them on callback?

uamqp/message.py
```diff
@@ -377,7 +377,7 @@ def accept(self):
             return True
         return False
 
-    def reject(self, condition=None, description=None):
+    def reject(self, condition=None, description=None, info=None):
         """Send a response disposition to the service to indicate that
         a received message has been rejected. If the client is running in PeekLock
         mode, the service will wait on this disposition. Otherwise it will
```
annatisch, line 383: Please update the docstring :)

src/message_receiver.pyx
yunhaoling: AFAIK, we only have `settle_rejected_message` for rejecting..
I don't see a separate code path.

If you're referring to `delivery_rejected` in `message.pyx`, we don't use it in python code. it's for constructing a delivery  reject object.

uamqp/message.py
yunhaoling: updated

uamqp/errors.py
KieranBrantnerMagee: Chiming in with my usual dumb question: If this is breaking the public API, even if we have confident it's safe, should we bump the minor version # instead of patch?

uamqp/message.py
```diff
@@ -389,13 +389,16 @@ def reject(self, condition=None, description=None):
         :type condition: bytes or str
         :param description: A description/reason to accompany the rejection.
         :type description: bytes or str
+        :param info: Information about the error condition.
+        :type info: dict
```
KieranBrantnerMagee, line 393: So we call it description/reason in SB; is there a reason for the nomenclature split, or should we normalize?

yunhaoling: yeah, it's the spec of `reject` in amqp, SB is utilizing the `amqp.reject` for `sb.dead-letter`.

```
-- Copied from AMQP

<type name="rejected" class="composite" source="list" provides="delivery-state, outcome">
    <descriptor name="amqp:rejected:list" code="0x00000000:0x00000025"/>
    <field name="error" type="error"/>
</type>


<type name="error" class="composite" source="list">
    <descriptor name="amqp:error:list" code="0x00000000:0x0000001d"/>
    <field name="condition" type="symbol" requires="error-condition" mandatory="true"/>
    <field name="description" type="string"/>
    <field name="info" type="fields"/>
</type>

Field Details
condition - error condition
    A symbolic value indicating the error condition.
description - descriptive text about the error condition
    This text supplies any supplementary details not indicated by the condition eld.
    This text can be logged as an aid to resolving issues.
info - map carrying information about the error condition
```

uamqp/errors.py
yunhaoling: `info` has been moved to the end making it non-breaking

PR #163, opened 2020-07-10
Keep receiver alive after timeout and reuse

yunhaoling:
In current world after timeout, the receiver would be shutdown, everything needs to be re-established.

- receiver life cycle control on timeout: added a new kwarg "_shutdown_after_timeout" into ReceiveClient Constructor to allow the upper layer control the life cycle
- after one generator ends, new one can simply be created to continue utilizing the receiver.

To reuse the receiver in the service bus, the code could be:
```python


class ServiceBusReceiver:
    def _create_handler(self, auth):
        self._handler = ReceiveClient(..., shutdown_after_timeout=False)

    def __next__(self):
        self._check_live()
        while True:
            try:
                return self._do_retryable_operation(self._iter_next)
            except StopIteration:
                self._message_iter = None 

    def _iter_next(self):
        self._open()
        if not self._message_iter:
            self._message_iter = self._message_iter = self._handler.receive_messages_iter()  # get a new generator if last one ends
        uamqp_message = next(self._message_iter)
        message = self._build_message(uamqp_message)
        return message

```

### DIFF
--- a/uamqp/async_ops/client_async.py
+++ b/uamqp/async_ops/client_async.py
@@ -875,6 +875,7 @@ class ReceiveClientAsync(client.ReceiveClient, AMQPClientAsync):
         await self.open_async()
         self._message_received_callback = on_message_received
         self._timeout_reached = False
+        self._last_activity_timestamp = None
         receiving = True
         try:
             while receiving and not self._timeout_reached:
@@ -884,7 +885,6 @@ class ReceiveClientAsync(client.ReceiveClient, AMQPClientAsync):
             raise
         finally:
             self._streaming_receive = False
-            self._timeout_reached = False
             if not receiving and self._shutdown_after_timeout:
                 await self.close_async()
 
@@ -929,26 +929,24 @@ class ReceiveClientAsync(client.ReceiveClient, AMQPClientAsync):
             return batch
 
         self._timeout_reached = False
-        try:
-            while receiving and not expired and len(batch) < max_batch_size and not self._timeout_reached:
-                while receiving and self._received_messages.qsize() < max_batch_size and not self._timeout_reached:
-                    if timeout and self._counter.get_current_ms() > timeout:
-                        expired = True
-                        break
-                    before = self._received_messages.qsize()
-                    receiving = await self.do_work_async()
-                    received = self._received_messages.qsize() - before
-                    if self._received_messages.qsize() > 0 and received == 0:
-                        # No new messages arrived, but we have some - so return what we have.
-                        expired = True
-                        break
+        self._last_activity_timestamp = None
+        while receiving and not expired and len(batch) < max_batch_size and not self._timeout_reached:
+            while receiving and self._received_messages.qsize() < max_batch_size and not self._timeout_reached:
+                if timeout and self._counter.get_current_ms() > timeout:
+                    expired = True
+                    break
+                before = self._received_messages.qsize()
+                receiving = await self.do_work_async()
+                received = self._received_messages.qsize() - before
+                if self._received_messages.qsize() > 0 and received == 0:
+                    # No new messages arrived, but we have some - so return what we have.
+                    expired = True
+                    break
 
-                while not self._received_messages.empty() and len(batch) < max_batch_size:
-                    batch.append(self._received_messages.get())
-                    self._received_messages.task_done()
-            return batch
-        finally:
-            self._timeout_reached = False
+            while not self._received_messages.empty() and len(batch) < max_batch_size:
+                batch.append(self._received_messages.get())
+                self._received_messages.task_done()
+        return batch
 
     def receive_messages_iter_async(self, on_message_received=None):
         """Receive messages by asynchronous generator. Messages returned in the
@@ -1030,6 +1028,5 @@ class AsyncMessageIter(collections.abc.AsyncIterator):  # pylint: disable=no-mem
             self.receiving = False
             raise
         finally:
-            self._client._timeout_reached = False  # pylint: disable=protected-access
             if not self.receiving and self._client._shutdown_after_timeout:
                 await self._client.close_async()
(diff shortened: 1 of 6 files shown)


uamqp/async_ops/client_async.py
```diff
@@ -840,8 +840,13 @@ async def _client_run_async(self):
             if self._timeout > 0:
                 timespan = now - self._last_activity_timestamp
                 if timespan >= self._timeout:
-                    _logger.info("Timeout reached, closing receiver.")
-                    self._shutdown = True
+                    self._timeout_reached = True
+                    if self._shutdown_after_timeout:
```
KieranBrantnerMagee, line 844: I don't see a docstring for this being passed in or defined anywhere?

yunhaoling: it's in client.py:
`self._shutdown_after_timeout = kwargs.pop('shutdown_after_timeout', True)`

TBH, I didn't plan to make this parameter public, because it seems to have a conflict with the current `timeout` param on the client:
`:param timeout: A timeout in milliseconds. The receiver will shut down if no
     new messages are received after the specified timeout. If set to 0, the receiver
     will never timeout and will continue to listen. The default is 0.
`

or we could make it public but add some additional notes in the `timeout`? like:
`:param timeout: A timeout in milliseconds. The receiver will shut down **if shutdown_after_timeout is set to true** and no
     new messages are received after the specified timeout. If set to 0, the receiver
     will never timeout and will continue to listen. The default is 0.
`

@annatisch do you think we shall make this parameter public?

uamqp/client.py
```diff
@@ -946,8 +949,13 @@ def _client_run(self):
...
                 timespan = now - self._last_activity_timestamp
                 if timespan >= self._timeout:
-                    _logger.info("Timeout reached, closing receiver.")
-                    self._shutdown = True
+                    self._timeout_reached = True
+                    if self._shutdown_after_timeout:
+                        _logger.info("Timeout reached, closing receiver.")
+                        self._shutdown = True
+                    else:
+                        self._last_activity_timestamp = None # To reuse the receiver, reset the timestamp
```
KieranBrantnerMagee, line 957: This comment is excellent (in terms of value added) but begs the question: How would one be expected to do this in normal operation?

(My understanding, to doubly use this to ensure I'm on the right page: ".  This can be achieved by performing a do_work loop until client_ready is true."  although feel free to be more terse if that can be summarized better.)

uamqp/async_ops/client_async.py
annatisch: Seems like it would be okay to make this public.

annatisch: My only though is that I'm not sure how well this plays with the other receive mechanisms. Though it makes sense for the iterator. I'm wondering - rather than make this a top-level client parameter - maybe we could only expose it at the iterator function?

yunhaoling: @annatisch This is a pretty good point of view which pushes me to think harder on this.

The right thing to do here from my point after revisiting the code base is: **It should be a top-level client parameter. And to your concern, we should ensure the parameter plays well with other receive mechanisms -- write tests for them.**

I should admit  the idea of only exposing it at the iterator looks the most simplistic. However:
  - Our goal is to keep the uamqp receiver alive in ServiceBus world which I don't think iterator only would help us achieve.
    - uamqp has three receiving mechanisms in total: `pull/push/iterator`.
    - Service bus uses `pull` and `iterator`.
      - Keeping uamqp receiver alive (not shutdown after timeout) should work for both `pull` and `iterator` mode.
      - If it's iterator only, then it means pull receive would shutdown the uamqp receiver after timeout.
  - Consistency concern: The key code to impact the behavior of "shutdown after timeout"  is in `do_work()` which is the base of all receiving mechanisms.
    - Making it generator only would bring extra complexity than it looks like, in which extra variables or flow need to be introduced to guarantee the consistency of receiver status when multiple receiving mechanisms are used simultaneously.

So overall, I believe making it a top-level parameter brings us more benefits than the costs as this meets our goal in SB and keeps the receiving flow clean and consistent in all ways.

uamqp/client.py
yunhaoling: I'm assuming "do this" in your question refers to `resetting self._last_activity_timestamp to None` and "normal operation" refers to all the receiving mechanisms Anna mentioned in the above comment.

so the answer to your question is that users doesn't have control over the resetting in current implementation, Timeout reaches in `do_work` is the only trigger to reset it.

But your understanding of the question makes me wonder if the "reset" logic should be more explicit -- just as how you understand, fetching a new receive_messages_iter would do the reset.

Another option would be to reset `self._last_activity_timestamp` proactively every time we get a new iterator, no longer triggered by timeout, and we need to this for all of our receiving mechanisms as mentioned by Anna.

yunhaoling: I need to write some code to feel which option is better and taking all of our mechanisms into consideration.
feel free to leave your thoughts.

uamqp/async_ops/client_async.py
yunhaoling: @annatisch , I have cover all the receiving mechanisms to work with new parameters and tests are added.

uamqp/client.py
yunhaoling: @KieranBrantnerMagee , I think you're right, I should be more explicit on this.
I've added "reset self._last_activity_timestamp" logic in each receive mechanism, matching your understanding now.

KieranBrantnerMagee: I should have responded back sooner to provide clarity, but I think this actually ended up for the best, because my head wasn't entirely fully "together' on what the correct comment should say, and your takeaway sounds far more well-thought-out than what I had in mind.  Will read the implementation with this context.

KieranBrantnerMagee:
Does this need a changelog note?

PR #171, opened 2020-09-23
message footer and annotation_delivery encoding fix

yunhaoling:
This PR fixes the issue https://github.com/Azure/azure-uamqp-python/issues/169 that footer and annotation_delivery not being encoded into the outgoing payload.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,8 +6,7 @@ Release History
 1.2.11 (Unreleased)
 +++++++++++++++++++
 
-- Fixed bug where `Message.footer` and `Message.delivery_annotation` not being encoded into the outgoing payload.
-
+- Fixed bug where `Message.footer` and `Message.delivery_annotation` were not encoded into the outgoing payload.
 
 1.2.10 (2020-08-05)
 +++++++++++++++++++
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,12 @@
 Release History
 ===============
 
+1.2.11 (Unreleased)
++++++++++++++++++++
+
+- Fixed bug where `Message.footer` and `Message.delivery_annotation` not being encoded into the outgoing payload.
+
+
 1.2.10 (2020-08-05)
 +++++++++++++++++++
 
--- a/samples/test_azure_event_hubs_receive.py
+++ b/samples/test_azure_event_hubs_receive.py
@@ -496,6 +496,76 @@ def test_event_hubs_dynamic_issue_link_credit(live_eventhub_config):
     log.info("Finished receiving")
 
 
+def test_event_hubs_send_event_with_amqp_attributes_sync(live_eventhub_config):
+    def data_generator():
+        for i in range(2):
+            msg = uamqp.message.Message(body='Data')
+            # header is only used on received msg, not set on messages being sent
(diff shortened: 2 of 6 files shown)


HISTORY.rst
```diff
@@ -3,6 +3,12 @@
 Release History
 ===============
 
+1.2.11 (Unreleased)
++++++++++++++++++++
+
+- Fixed bug where `Message.footer` and `Message.delivery_annotation` not being encoded into the outgoing payload.
```
KieranBrantnerMagee, line 9:
```suggestion
- Fixed bug where `Message.footer` and `Message.delivery_annotation` were not being encoded into the outgoing payload.
```
(small grammar nit)

samples/test_azure_event_hubs_receive.py
```diff
@@ -496,6 +496,77 @@ def message_received_callback(message):
...
+    receive_client = uamqp.ReceiveClient(source, auth=sas_auth, timeout=5000, debug=False, prefetch=10)
+    gen = receive_client.receive_messages_iter()
+    for message in gen:
+        if message.application_properties and message.application_properties.get(b'msg_type'):
+            if not rich_single_received:
+                rich_single_received = message.application_properties.get(b'msg_type') == b'rich_single'
+            if not rich_batch_received:
+                rich_batch_received = message.application_properties.get(b'msg_type') == b'rich_batch'
+            assert message.properties.message_id == b'richid'
+            assert message.delivery_annotations
```
annatisch, line 553: Is there anything in the spec on whether delivery_annotations _supposed_ to be encoded on outgoing payload? I'm curious whether these are only intended to be written to by the broker? Probably not, but still curious :)

annatisch (approved):
LGTM 👍

samples/test_azure_event_hubs_receive.py
yunhaoling: there doesn't seem to be any limitation according to the spec, and based on the test, the service supports setting these two properties :)

---

**delivery annotation spec**: http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-delivery-annotations

> The delivery-annotations section is used for delivery-specific non-standard properties at the head of the message. **Delivery annotations convey information from the sending peer to the receiving peer**. If the recipient does not understand the annotation it cannot be acted upon and its effects (such as any implied propagation) cannot be acted upon. Annotations might be specific to one implementation, or common to multiple implementations. The capabilities negotiated on link attach and on the source and target SHOULD be used to establish which annotations a peer supports. A registry of defined annotations and their meanings is maintained [AMQPDELANN]. The symbolic key "rejected" is reserved for the use of communicating error information regarding rejected messages. Any values associated with the "rejected" key MUST be of type error.
If the delivery-annotations section is omitted, it is equivalent to a delivery-annotations section containing an empty map of annotations.

**footer spec**: http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-footer
> The footer section is used for details about the message or delivery which can only be calculated or evaluated once the whole bare message has been constructed or seen (for example message hashes, HMACs, signatures and encryption details).
A registry of defined footers and their meanings is maintained [AMQPFOOTER].

PR #193, opened 2020-12-16
Sub module update

yunhaoling:
divide and conquer:

step1: remove the old azure-uamqp-c
step2: copy and paste the latest azure-uamqp-c (recursive)
step3: fix compiling errors
step4: revert each commit that happened in vendored directory and make PR to the c repo
step5: clean up useless c dependencies -- don't need to vendor all of them as they might be included in the final built wheels- 
- azure-ctest -- can remove
- testsiwthcer -- can remove
- umock -- we need this one now

step6: livetest and stresstest

---------- fix -----------

to compile the latest c module:
- stdbool/stdint.h missing
  - vendor and include it in the setup.py
  - there could be another way, just include the `u_mock/aux_inc` in CMakeLists.txt
- Fix C89 compilation error
  - Syntax fix
    - src/vendor/azure-uamqp-c/deps/azure-c-shared-utility/adapters/x509_schannel.c
    - src/vendor/azure-uamqp-c/deps/umock-c/src/umocktypes_wcharptr.c
    - src/vendor/azure-uamqp-c/deps/azure-c-shared-utility/src/constbuffer_array_batcher.c
  - Change compilation warning level from W4 to W3 on MSVC to address warning treated as error
    - Method "CertVerifyCertificateChainPolicy" (etc.) causes warning C4306 type case: conversion from int to LPCSTR, and this compilation warning is treated as error
  - Exclude `src_lock` module which is not used by any module but causing trouble to compiler
    - It calls "interlocked<XXX>64" operations which are not able to be compiled with VC9 X86
    - check here for more knowledge: https://github.com/idealvin/co/issues/38
- Windows SDK on Python2.7
  - In setup.py, utilize "get_latest_windows_sdk" to find the latest shared include dir (which includes header files)
  - In "get_generator_flags", append command to tell Cmake whether to use the latest WinSDK and Where is the include
  - In CMakeLists.txt, check flag and include directory


yunhaoling:
use PR:https://github.com/Azure/azure-uamqp-python/pull/200 which has cleaner commit history

PR #200, opened 2021-01-21
Update azure-uamqp-c submodule

yunhaoling:
echoing PR: https://github.com/Azure/azure-uamqp-python/pull/193 with more clean commit history

to compile the latest c module:
- stdbool/stdint.h missing
  - vendor and include it in the setup.py
- Fix C89 compilation error
  - Syntax fix
    - src/vendor/azure-uamqp-c/deps/azure-c-shared-utility/adapters/x509_schannel.c
    - src/vendor/azure-uamqp-c/deps/umock-c/src/umocktypes_wcharptr.c
    - src/vendor/azure-uamqp-c/deps/azure-c-shared-utility/src/constbuffer_array_batcher.c
  - Change compilation warning level from W4 to W3 on MSVC to address warning treated as error
    - Method "CertVerifyCertificateChainPolicy" (etc.) causes warning C4306 type case: conversion from int to LPCSTR, and this compilation warning is treated as error
  - Exclude `srw_lock` module which is not used by any module but causing trouble to compiler
    - It calls "interlocked<XXX>64" operations which are not able to be compiled with VC9 X86
- Windows SDK on Python2.7
  - In setup.py, utilize "get_latest_windows_sdk" to find the latest shared include dir (which includes header files)
  - In "get_generator_flags", append command to tell Cmake whether to use the latest WinSDK and Where is the include
  - In CMakeLists.txt, check flag and include directory


annatisch (approved):
LGTM :)
Awesome work!!!

lmazuel (approved):
Just looked at the setup.py and it looks good :)

PR #201, opened 2021-01-25
Fix memory leak when creating cbs token string

yunhaoling:


src/cbs.pyx
```diff
@@ -32,9 +32,9 @@ cpdef create_sas_token(const char* key, const char* scope, const char* keyname,
         raise ValueError("Failed to create SAS token.")
     if c_utils.SASToken_Validate(str_value) != True:
         raise ValueError("Generated invalid SAS token")
-    return c_strings.STRING_c_str(str_value)
-    #c_strings.STRING_delete(str_value)
-    #return c_string
```
annatisch, line 37: There's also line 28

src/cbs.pyx
```diff
@@ -32,9 +32,9 @@ cpdef create_sas_token(const char* key, const char* scope, const char* keyname,
...
+    cdef bytes py_str = c_strings.STRING_c_str(str_value)
+    c_strings.STRING_delete(str_value)
+    return py_str
```
annatisch, line 37: I've had trouble with this before - make sure you run this thoroughly through the 72 hour tests with high stress (e.g. high number of processes)

yunhaoling: do you remember the kind of trouble briefly? seg fault or invalid string value etc.

yunhaoling:
3-days long running test passed, looks good

PR #219, opened 2021-03-24
add support for sending/receiving message of amqp sequence type

yunhaoling:
- Based on the option1 design proposal: https://gist.github.com/yunhaoling/b7c30894f59542c70de589940a319460
- Added support for AMQP Sequence as the body type of the message.
- Added new class `uamqp.MessageBodyType` to denote the body type of an amqp message, including:
  - `DataType`: The body consists of one or more data sections and each section contains opaque binary data.
  - `SequenceType`: The body consists of one or more sequence sections and each section contains an arbitrary number of structured data elements.
  - `ValueType`: The body consists of one amqp-value section and the section contains a single AMQP value.
- Added new parameter `body_type` to the constructor of `uamqp.Message` which takes `uamqp.MessageBodyType` to denote the body type of an amqp message.
- Added support for sending/receiving message whose body consists of sequence into the c uamqp library

### DIFF
--- a/src/vendor/azure-uamqp-c/src/message_sender.c
+++ b/src/vendor/azure-uamqp-c/src/message_sender.c
@@ -207,6 +207,7 @@ static SEND_ONE_MESSAGE_RESULT send_one_message(MESSAGE_SENDER_INSTANCE* message
         AMQP_VALUE application_properties_value = NULL;
         AMQP_VALUE body_amqp_value = NULL;
         size_t body_data_count = 0;
+        size_t body_sequence_count = 0;
         AMQP_VALUE msg_annotations = NULL;
         AMQP_VALUE footer = NULL;
         AMQP_VALUE delivery_annotations = NULL;
@@ -442,6 +443,62 @@ static SEND_ONE_MESSAGE_RESULT send_one_message(MESSAGE_SENDER_INSTANCE* message
                 }
                 break;
             }
+
+            case MESSAGE_BODY_TYPE_SEQUENCE:
+            {
+                AMQP_VALUE message_body_amqp_sequence;
+                size_t i;
+
+                if (message_get_body_amqp_sequence_count(message, &body_sequence_count) != 0)
+                {
+                    LogError("Cannot get body AMQP sequence count");
+                    result = SEND_ONE_MESSAGE_ERROR;
+                }
+                else
+                {
+                    if (body_sequence_count == 0)
+                    {
+                        LogError("Body sequence count is zero");
+                        result = SEND_ONE_MESSAGE_ERROR;
+                    }
+                    else
+                    {
+                        for (i = 0; i < body_sequence_count; i++)
+                        {
+                            if (message_get_body_amqp_sequence_in_place(message, i, &message_body_amqp_sequence) != 0)
+                            {
+                                LogError("Cannot get body AMQP sequence %u", (unsigned int)i);
+                                result = SEND_ONE_MESSAGE_ERROR;
+                            }
+                            else
+                            {
+                                AMQP_VALUE body_amqp_sequence;
+                                body_amqp_sequence = amqpvalue_create_amqp_sequence(message_body_amqp_sequence);
+                                if (body_amqp_sequence == NULL)
+                                {
+                                    LogError("Cannot create body AMQP sequence");
+                                    result = SEND_ONE_MESSAGE_ERROR;
+                                }
+                                else
+                                {
+                                    if (amqpvalue_get_encoded_size(body_amqp_sequence, &encoded_size) != 0)
+                                    {
+                                        LogError("Cannot get body AMQP sequence encoded size");
+                                        result = SEND_ONE_MESSAGE_ERROR;
+                                    }
+                                    else
+                                    {
+                                        total_encoded_size += encoded_size;
+                                    }
+
+                                    amqpvalue_destroy(body_amqp_sequence);
+                                }
+                            }
+                        }
+                    }
+                }
+                break;
+            }
             }
 
             if (result == 0)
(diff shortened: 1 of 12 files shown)


uamqp/constants.py
```diff
@@ -170,3 +170,21 @@ class TransportType(Enum):
     """
     Amqp = 1
     AmqpOverWebsocket = 2
+
+
+class MessageBodyType(Enum):
```
annatisch, line 175: The spec refers to this as "message format" - so maybe we should call it `MessageBodyFormat`?

uamqp/constants.py
```diff
@@ -170,3 +170,21 @@ class TransportType(Enum):
...
+     number of structured data elements.
+    `ValueType`: The body consists of one amqp-value section and the section contains a single AMQP value.
+
+    Please refer to the AMQP spec:
+    http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#section-message-format
+    for further information on the message body type.
+    """
+    DataType = c_uamqp.MessageBodyType.DataType
+    ValueType = c_uamqp.MessageBodyType.ValueType
+    SequenceType = c_uamqp.MessageBodyType.SequenceType
```
annatisch, line 190: Likewise - I think we can drop the "type" suffix:
```
MessageBodyFormat.Data
MessageBodyFormat.Value
MessageBodyFormat.Sequence
```

annatisch: Also - what are the "real" values of these? Are they integers?

uamqp/message.py
```diff
@@ -44,6 +48,14 @@ class Message(object):
     :type header: ~uamqp.message.MessageHeader
     :param msg_format: A custom message format. Default is 0.
     :type msg_format: int
+    :param body_type: The AMQP body type used to specify the type of the body section of an amqp message.
```
annatisch, line 51: Wondering if `body_format` is a better name....

uamqp/message.py
```diff
@@ -252,7 +260,7 @@ def _parse_message_body(self, message):
         elif body_type == c_uamqp.MessageBodyType.DataType:
             self._body = DataBody(self._message)
         elif body_type == c_uamqp.MessageBodyType.SequenceType:
-            raise TypeError("Message body type Sequence not supported.")
+            self._body = SequenceBody(self._message)
```
annatisch, line 263: Looks like the constructors of these Body objects were already set up to take the c message object - so it's a good thing not to expose these objects directly.

uamqp/message.py
```diff
@@ -70,6 +82,7 @@ def __init__(self,
                  message=None,
                  settler=None,
                  delivery_no=None,
+                 body_type=None,
```
annatisch, line 85: Wouldn't this be a breaking change to the constructor?

uamqp/message.py
```diff
@@ -264,6 +272,57 @@ def _can_settle_message(self):
...
+        if isinstance(body, (six.text_type, six.binary_type)) or \
+                (isinstance(body, list) and all([isinstance(b, (six.text_type, six.binary_type)) for b in body])):
+            return constants.MessageBodyType.DataType
+        return constants.MessageBodyType.ValueType
+
+    @staticmethod
+    def _validate_body_and_body_type(body, body_type):
+        if body_type == constants.MessageBodyType.DataType:
+            if not (isinstance(body, (six.text_type, six.binary_type)) or
+                    (isinstance(body, list) and all([isinstance(b, (six.text_type, six.binary_type)) for b in body]))):
```
annatisch, line 291: Seems a shame to have to do this if-statement twice for what is currently the dominant code path. Do you think it could have a perf impact for high volumes of messages?

uamqp/message.py
```diff
@@ -264,6 +272,57 @@ def _can_settle_message(self):
...
+            pass
+        else:
+            raise ValueError("Unsupported MessageBodyType: {}".format(body_type))
+
+    def _set_body_by_body_type(self, body, body_type):
+        if body_type == constants.MessageBodyType.DataType:
+            self._body = DataBody(self._message)
+            if isinstance(body, (six.text_type, six.binary_type)):
+                self._body.append(body)
+            elif isinstance(body, list) and all([isinstance(b, (six.text_type, six.binary_type)) for b in body]):
```
annatisch, line 312: Make that 3 times that we do these if statements ;)

src/vendor/azure-uamqp-c/src/message_sender.c
```diff
@@ -442,6 +443,62 @@ static SEND_ONE_MESSAGE_RESULT send_one_message(MESSAGE_SENDER_INSTANCE* message
                 }
                 break;
             }
+
+            case MESSAGE_BODY_TYPE_SEQUENCE:
```
annatisch, line 447: It might be a good idea to run some cross language tests to ensure that the "sequence body" messages that we're generating here are recognized and consumable by a different AMQP implementation.

annatisch: (And in reverse)

uamqp/constants.py
yunhaoling: I believe the term message format is reserved for the whole structure of a message, not just the body.

> 3.2.16 MESSAGE-FORMAT
MESSAGE-FORMAT: the format + revision for the messages fined  by this document 0
This value goes into the message-format field of the transfer frame when transferring messages of the
format defined herein.

> The exact structure of a message, together with its encoding, is defined by the message format. This
document defines the structure and semantics of message format 0 (MESSAGE-FORMAT). Altogether
a message consists of the following sections

e.g. the message is composed of header, delivery-annotations, properties, body, etc.

so we should keep the name type..

uamqp/constants.py
yunhaoling: +1 on dropping the `Type` suffix

uamqp/message.py
yunhaoling: see comment on the format vs type :P

uamqp/message.py
yunhaoling: good catch, yes, I should move this to the end of the constructor to avoid breaking change

src/vendor/azure-uamqp-c/src/message_sender.c
yunhaoling: agreed, we'll do a cross-language test

uamqp/message.py
yunhaoling: code updated

annatisch (approved):

PR #222, opened 2021-03-26
add setstate and getstate to message for pickling

swathipil:


uamqp/message.py
```diff
@@ -116,6 +120,42 @@ def __init__(self,
...
+        state["_body"] = list(state["_body"].data) if state["_body"].data else ""
+
+        state["state"] = None
+        state["delivery_no"] = None
+        state["delivery_tag"] = None
+        # need _response for settled property
+        state["_response"] = self.settled
+        state["_settler"] = None
+        state["on_send_complete"] = None
+        state["_message"] = None
```
yunhaoling, line 134: I'm wondering setting them to None is not enough, or it should be `del state[key]`.

Do you have an sample of the pickled bytes of a message object?

This also make me think that we probably should echo what we did for MessageHeader/MessageProperties: creating an empty dic and then pick the properties that we care.

uamqp/message.py
```diff
@@ -116,6 +120,42 @@ def __init__(self,
             self._annotations = annotations
             self._header = header
 
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        state["_body"] = list(state["_body"].data) if state["_body"].data else ""
```
yunhaoling, line 125: things will be tricky after I add the sequence type support.
I'll take care of this part.

uamqp/message.py
```diff
@@ -116,6 +120,42 @@ def __init__(self,
...
+            self._body.append(body)
+        elif isinstance(body, list) and all(
+            [isinstance(b, (six.text_type, six.binary_type)) for b in body]
+        ):
+            self._body = DataBody(self._message)
+            for value in body:
+                self._body.append(value)
+        else:
+            self._body = ValueBody(self._message)
+            self._body.set(body)
```
yunhaoling, line 153: I'll take care of handling different body types 😉

uamqp/message.py
yunhaoling, line 125: also wondering in the state dict, should we have key with the underscore which indicates it's a private thing?
or we should just have `state['body']=...`.

from my perspective, pickle support is part of the public api surface, and by setting `_<variable>` then we're leaking kinda private information.

tests/test_message.py
```diff
@@ -29,3 +30,55 @@ def test_message_proeprties():
...
+
+    header = MessageHeader()
+    header.delivery_count = 3
+    header.time_to_live = 5
+    header.first_acquirer = 'dkfj'
+    header.durable = True
+    header.priority = 4
+
+
+    message = Message(properties=properties, header=header)
```
yunhaoling, line 58: let's also add tests for data different body types -- it's on me

uamqp/message.py
```diff
@@ -727,21 +806,33 @@ def __init__(self,
...
+                "creation_time": self.creation_time,
+                "group_id": self.group_id,
+                "group_sequence": self.group_sequence,
+                "reply_to_group_id": self.reply_to_group_id,
+            }
+        )
+
+    def __getstate__(self):
+        state = self._get_properties_dict()
+        state["_encoding"] = self._encoding
```
yunhaoling, line 829: I guess pickle should have its own encode format support? we probably don't need to include our encoding information in the state.

uamqp/message.py
swathipil: So I decided to set the `state[key] = None` rather than deleting them b/c, for instance, if a user tries to pickle an event and repickle it, deleting the key results in a `KeyError`. for example:
```
 pickle_ed = pickle.loads(pickle.dumps(ed))
 repickle_ed = pickle.loads(pickle.dumps(pickle_ed))   # results in KeyError
```
When you say sample of the pickled bytes, do you mean maybe a pickle file of a pickled message object?

Makes sense. I believe the properties that would require getstate/setstate for their respective types are _message, _settler, on_send_complete, and state. The `cMessage` class for `_message` is the only one I don't think I should touch, since it deals with the c files??

uamqp/message.py
swathipil: that makes sense, and I'm assuming that should be the case for all the private attributes?

uamqp/message.py
swathipil: This encoding is the attribute associated with MessageProperties. I'm including it here because, in `__setstate__`, I reset the encoding of the MessageProperties object because it is passed into the `utils.data_factory(value, encoding=self._encoding)` call when resetting all of the attributes. I thought this encoding was separate from the pickle encoding, but my understanding of encoding might be wrong?

uamqp/message.py
yunhaoling: based on our discussion, in __getstate__ we pick a minimum set of objects to serialize:
- delivery_no
- delivery_tag
- properties
- application_properties 
- annotations
- header
- footer
- delivery_annotations
- _body
- _body_type
(feel free to call out if there's any other private objects that are required to be encoded)
so we would be doing:

```python
def __getstate__(self):
   state = {}
   state["application_properties"] = self.application_properties
   ...
   return state
```

as far as I could see, this gives us two benefits:

1. extensibility, in the future we might want to add/remove instance variables
2. save encoded bytes: useless objects will not be encoded to save bytes

yunhaoling: and in the __setstate__, we manually populate the instance variables by default value which are not encoded:
```python
    def __setstate__(self, state):
        state["state"] = constants.MessageState.WaitingToBeSent
        state["idle_time"] = ...
        ...
        self.__dict__.update(state)
        # populate body
```

swathipil: that makes sense to me! One more property that is relevant is `settled`, which is determined by `_response`. To only expose `settled`, I plan to do something like the following:
```
def __getstate__(self):
    ....
    state['settled'] = self.settled  # True/False
    ...
def __setstate__(self, state):
    state['_response'] = errors.MessageAccepted()  if state['settled'] else None # arbitrary value for settled to be True

Comment (yunhaoling) [reply uamqp/message.py:155]: I don't think we could simply set state['_response'] to errors.MessageAccepted() based on `settled` is True or not.
because self._response could be
`MessageAlreadySettled`
`MessageAccepted`
`MessageRejected`
`MessageReleased`
`MessageModified`

depending on how `self._response` is used in current message (mostly for checking whether a message could be settled), then the public settled property is sufficient.

so I would keep `state['_response'] = None` for now to avoid introducing complexity.

Comment (swathipil) [reply uamqp/message.py:155]: since `def settled` method doesn't have a corresponding setter method, and it relies on `_response` for a return value, wouldn't we need to also serialize `_response`? If we set it to none, then even if `message.settled` was True before pickling, it would be False after.

File: uamqp/errors.py [hunk 6]
```diff
@@ -259,6 +259,9 @@ def __init__(self):
         response = "Invalid operation: this message is already settled."
         super(MessageAlreadySettled, self).__init__(response)
 
+    def __reduce__(self):
+        return (self.__class__, ())
+
```
yunhaoling, line 264: do we need the override here? seems like it's doing nothing..

uamqp/errors.py
```diff
@@ -282,6 +287,9 @@ def __init__(self, condition=None, description=None, encoding='UTF-8', info=None
         self.error_info = utils.data_factory(info, encoding=encoding) if info else None
         super(MessageRejected, self).__init__()
 
+    def __reduce__(self):
+        return (self.__class__, (self.error_condition, self.error_description, self._encoding, self._info))
```
yunhaoling, line 291: this is neat! nice work!

uamqp/message.py
```diff
@@ -1235,13 +1334,32 @@ def __init__(self, header=None):
...
+            self.__setattr__(key, val)
+
+    def _get_properties_dict(self):
+        return {
+            "delivery_count": self.delivery_count,
+            "time_to_live": self.time_to_live,
+            "first_acquirer": self.first_acquirer,
+            "durable": self.durable,
+            "priority": self.priority,
+        }
```
yunhaoling, line 1362: wondering do we really need those for MessageHeader.. all the ivars are python objects

uamqp/errors.py
swathipil: for some reason when this was removed,  the following results in an error:
```
    internal_message = c_uamqp.create_message()
    internal_message.add_body_data(b"hi")
    message_w_message_param = Message(
        message=internal_message,
        delivery_no=1
    )
```

The error: 
```
E       TypeError: __init__() takes 1 positional argument but 2 were given

```

yunhaoling (approved):
LGTM

PR #233, opened 2021-04-24
add EH/SB tests to livetest pipeline

swathipil:
fixes: https://github.com/Azure/azure-sdk-for-python/issues/11588


.azure-pipelines/client.test.live.yml
```diff
@@ -311,3 +310,590 @@ jobs:
...
+      - script: |
+          python --version
+          python -m pip install pytest
+          echo "Installing $DownloadArtifactWhl"
+          python -m pip install $(Pipeline.Workspace)/$(DownloadArtifactFolder)/$(DownloadArtifactWhl)
+          cd $(SdkPath)
+          python -m pip install --ignore-installed .
+          python -m pip install -r dev_requirements.txt
+          python -m pip list
+          pytest tests -v --doctest-modules --ignore=tests/perfstress_tests --junitxml=junit/eventhub/azure-eventhub/test-results-live.xml
```
scbedd, line 418: Don't abandon `python -m`! Everything is right up till here :D

scbedd (changes requested):
All these jobs run pretty much exactly the same command. I think we you see where you have `strategy: ` set? And there is a single entry? I think we can put _all_ of these into the same single job definition that will be multiplied by the strategy alone. Can you point out anything that's honestly different between these runs that can't be covered by the variable? 

For your first attempt I'd just merge the matrix entries into a single one, and see what happens!  You can remove `MaxParallel` if you don't want them to run in serial :) 

Other than that single (tbh a bit of work) feedback, this looks good!

.azure-pipelines/client.test.live.yml
```diff
@@ -311,3 +310,590 @@ jobs:
...
+      DevReqsPath: './azure-sdk-for-python/sdk/eventhub/azure-eventhub/dev_requirements.txt'
+      DownloadArtifactFolder: 'uamqp-win$(PythonArchitecture)-$(PythonVersion)-whl'
+      SdkPath: 'azure-sdk-for-python/sdk/eventhub/azure-eventhub'
+    steps:
+      - template: /.azure-pipelines/use-python-version.yml
+        parameters:
+          architecture: '$(PythonArchitecture)'
+          versionSpec: '$(PythonVersion)'
+      - script: |
+          git clone https://github.com/Azure/azure-sdk-for-python.git --single-branch
```
scbedd, line 337: We want something like a `sparse` checkout here.  I _think_ you can just add `-–depth 1` to your current command.

.azure-pipelines/client.test.live.yml
```diff
@@ -311,3 +310,590 @@ jobs:
...
+        displayName: 'Clone azure-sdk-for-python'
+      - task: DownloadPipelineArtifact@2
+        displayName: 'Downloading artifact'
+        inputs:
+          source: 'current'
+          patterns: '$(DownloadArtifactFolder)/*.whl'
+      - pwsh: |
+          $whlfile = Get-ChildItem -Filter *.whl $(Pipeline.Workspace)/$(DownloadArtifactFolder) | Select-Object -First 1 -ExpandProperty Name
+          echo "##vso[task.setvariable variable=DownloadArtifactWhl]$whlfile"
+          dir $(Pipeline.Workspace)/$(DownloadArtifactFolder)
```
scbedd, line 347: When you're running `pwsh`, `dir` is just aliased `Get-ChildItem`.

image

.azure-pipelines/client.test.live.yml
```diff
@@ -311,3 +310,590 @@ jobs:
...
+        displayName: 'Get name of whl file to install'
+      - script: |
```
scbedd, line 349: NIT: is there a reason we're swapping between `pwsh` and `script` here?

Not that the commands really matter, but it's good to keep consistency.

.azure-pipelines/client.test.live.yml
scbedd: (this same feedback is present for all the job definitions)

.azure-pipelines/client.test.live.yml
swathipil: I just didn't want to think about the logic for script, so just copy-pasted the pwsh commands from a previous job :P
It's prob not too hard to translate so will fix 👍

.azure-pipelines/client.yml
```diff
@@ -382,4 +382,4 @@ jobs:
         displayName: 'Publish wheel artifacts'
         inputs:
           artifactName: uamqp-$(Platform)-$(PythonVersion)-x64-whl
-          pathToPublish: 'wheelhouse'
+          pathToPublish: 'wheelhouse'
```
scbedd, line 385: nit: leave trailing newline! Good git etiquette

.azure-pipelines/test-eh-sb-sdk.yml
```diff
@@ -0,0 +1,59 @@
...
+      AZURE_CLIENT_ID: $(aad-azure-sdk-test-client-id)
+      AZURE_CLIENT_SECRET: $(aad-azure-sdk-test-client-secret)
+      AZURE_TEST_RUN_LIVE: 'true'
+  - task: PublishTestResults@2
+    displayName: 'Publish test results'
+    condition: succeededOrFailed()
+    inputs:
+      testResultsFiles: '**/$(TestResultsPath)'
+      testResultsFormat: 'JUnit'
+      testRunTitle: '$(SDK) Python $(PythonVersion)'
```
scbedd, line 59: Nit: we probably want to include the OS Arch here right? (Windows/Linux/MacOs) given we are running this for all of them right?

scbedd (approved):
I like it a lot! Remember to uncomment the strategies before you merge!

.azure-pipelines/test-eh-sb-sdk.yml
swathipil: ah yes, will do!

PR #241, opened 2021-06-15
[AMQP Python] Send single message and bug fix in decode to handle list32

yunhaoling:
todo: adding encode/decode test for list32

### DIFF
--- a/uamqp/_decode.py
+++ b/uamqp/_decode.py
@@ -276,14 +276,15 @@ def decode_frame(data):
     frame_type = data[2]
     compound_list_type = data[3]
     buffer = None
-    if compound_list_type == 0xc0:  # list8 0xc0
-        count = data[5]  # data[4] is size, data[5] is count
-        buffer = data[6:]
-    elif compound_list_type == 0xd0:  # list32 0xd0
-        count = c_signed_int.unpack(data[8:12])[0]  # data[4:8] is size, data[8:12] is count
+    compound_list_type = data[3]
+    if compound_list_type == 0xd0:
+        # list32 0xd0: data[4:8] is size, data[8:12] is count
+        count = c_signed_int.unpack(data[8:12])[0]
         buffer = data[12:]
-    else:  # list0 0x45
-        count = 0
+    else:
+        # list8 0xc0: data[4] is size, data[5] is count
+        count = data[5]
+        buffer = data[6:]
     fields = [None] * count
     for i in range(count):
         buffer, fields[i] = _DECODE_BY_CONSTRUCTOR[buffer[0]](buffer[1:])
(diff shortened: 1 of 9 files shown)


uamqp/client.py
```diff
@@ -237,6 +262,99 @@ def do_work(self, **kwargs):
...
+        on_send_complete = partial(self._on_send_complete, message_delivery)
+        delivery = self._link.send_transfer(
+            message_delivery.message,
+            on_send_complete=on_send_complete,
+            timeout=timeout
+        )
+        if not delivery.sent:
+            raise RuntimeError("Message is not sent.")
+
+    def _on_send_complete(self, message_delivery, message, reason, state):
```
annatisch, line 326: If I recall correctly - this callback can also be called if the message expires or the link goes down. Though we can probably add handling for that as needed.

uamqp/client.py
```diff
@@ -237,6 +262,99 @@ def do_work(self, **kwargs):
...
+        if SEND_DISPOSITION_ACCEPT in state:
+            message_delivery.state = MessageDeliveryState.Ok
+        else:
+            # TODO: sending disposition state could only be rejected/accepted?
+            message_delivery.state = MessageDeliveryState.Error
+            message_delivery.reason = reason
+
+    def send_message(self, message, timeout=0, **kwargs):
```
annatisch, line 334: We could roll timeout into kwargs

uamqp/_decode.py
```diff
@@ -274,9 +274,16 @@ def decode_frame(data):
...
-    # Ignore the next 2 bytes, they will always be list constructor, size.
-    count = data[5]
-    buffer = data[6:]
+    compound_list_type = data[3]
+    buffer = None
+    if compound_list_type == 0xc0:  # list8 0xc0
+        count = data[5]  # data[4] is size, data[5] is count
+        buffer = data[6:]
+    elif compound_list_type == 0xd0:  # list32 0xd0
+        _, count = _decode_int_large(data[8:12])  # data[4:8] is size, data[8:12] is count
```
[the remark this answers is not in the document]
yunhaoling: thanks for the suggestion, updated

uamqp/client.py
```diff
@@ -6,27 +6,49 @@
...
+from uamqp import constants
 from .error import AMQPConnectionError
 from .mgmt_operation import MgmtOperation
 
+
 _logger = logging.getLogger(__name__)
 _MAX_FRAME_SIZE_BYTES = 64 * 1024
 
 
+class _MessageDelivery:
```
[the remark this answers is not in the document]
yunhaoling: I agreed, initially put it here just for convenience :P

moved to message.py

uamqp/client.py
yunhaoling: yep, could do that.
just moved

uamqp/client.py
yunhaoling: added a "TODO" comment in the callback so that we don't forget to revisit this part in the future

PR #242, opened 2021-06-17
Client integration with CBS Authentication

yunhaoling:


uamqp/client.py
```diff
@@ -218,6 +234,11 @@ def auth_complete(self):
 
         :rtype: bool
         """
+        if self._cbs_authenticator:
+            if self._cbs_authenticator.state.value != 2 or not self._cbs_authenticator.handle_token():
```
yunhaoling, line 238: move the `_cbs_authenticator` check into the `handle_token` part? or at least improve the cbs_authenticator state to use Enums:
```
class CbsAuthenticatorState(Enum):
    Ok = 1
    Error = ...
```

uamqp/aio/_link_async.py
```diff
@@ -236,4 +236,4 @@ async def detach(self, close=False, error=None):
             await self._set_state(LinkState.DETACHED)
         elif self.state == LinkState.ATTACHED:
             await self._outgoing_detach(close=close, error=error)
-            await self._set_state(LinkState.ATTACH_SENT)
+            await self._set_state(LinkState.DETACH_SENT)
```
annatisch, line 239: haha - nice catch.

test_perf.py
```diff
@@ -82,7 +82,7 @@ async def client_receive_async(args, client, partition):
 
 def create_and_open_receive_client(args, partition, clients_arr=None):
     print("Creating and opening receive client:{}".format(partition))
-    creds = SASLPlainCredential(authcid=live_eventhub_config['key_name'], passwd=live_eventhub_config['access_key'])
+    creds = SASLPlainAuth(authcid=live_eventhub_config['key_name'], passwd=live_eventhub_config['access_key'])
```
annatisch, line 85: How do you feel about 'Credential' vs 'Auth' here?
Not a topic that needs to be discussed yet - as we can solidify naming much later. Just wondering if you have strong preferences.

uamqp/authentication.py
```diff
@@ -0,0 +1,177 @@
...
+        expires_on
+    )
+
+
+class SASLPlainAuth(object):
+    def __init__(self, authcid, passwd, authzid=None):
+        self.sasl = SASLPlainCredential(authcid, passwd, authzid)
+
+
+class _CBSAuth(object):
```
annatisch, line 50: Can we add a string literal discriminator for these `Auth` objects?

uamqp/authentication.py
```diff
@@ -0,0 +1,177 @@
...
+        return expires_in, expires_on
+
+
+class JWTTokenAuth(_CBSAuth):
+    def __init__(
+        self,
+        uri,
+        audience,
+        get_token,
+        token_type=TOKEN_TYPE_JWT,
```
annatisch, line 104: Let's roll all the optional parameters for all these auth types into `kwargs`.

uamqp/cbs.py
```diff
@@ -18,29 +18,27 @@
...
-            auth_audience,
-            get_token,
-            token_type=TOKEN_TYPE_SASTOKEN,
-            auth_timeout=AUTH_TIMEOUT,
-            encoding='UTF-8'
+        self,
+        session,
+        auth,
+        auth_timeout=AUTH_TIMEOUT,
+        encoding='UTF-8'
```
annatisch, line 41: I think we can strip out the encoding support. I believe CBS should always be utf-8, we can simply the code by simply removing this. Or not decode at all.

uamqp/cbs.py
```diff
@@ -51,18 +49,19 @@ def __init__(
...
-        if not get_token or not callable(get_token):
+        if not auth.get_token or not callable(auth.get_token):
             raise ValueError("get_token must be a callable object.")
 
-        self._get_token = get_token
-        self._token_type = token_type
+        self._auth = auth
+        self._audience = self._auth.audience
+        self._get_token = self._auth.get_token
+        self._token_type = self._auth.token_type
```
annatisch, line 59: Do we need to store these in local variables when we've already stored `self._auth`?

uamqp/client.py
```diff
@@ -75,6 +77,8 @@ class AMQPClient(object):
     :param idle_timeout: Timeout in milliseconds after which the Connection will close
      if there is no further activity.
     :type idle_timeout: int
+    :param auth_timeout: Timeout in seconds for authentication
```
annatisch, line 80: Maybe clarify that this is the timeout for token negotiation in the case of CBS auth? It will otherwise be ignored, right?

uamqp/client.py
```diff
@@ -188,6 +194,13 @@ def open(self):
             outgoing_window=self._outgoing_window
         )
         self._session.begin()
+        if isinstance(self._auth, _CBSAuth):
```
annatisch, line 197: If we add that string discriminator to the auth types, we wont have to do `isinstance` checks :)

uamqp/client.py
```diff
@@ -116,6 +120,8 @@ def __init__(self, hostname, auth=None, **kwargs):
         self._link = None
         self._socket_timeout = False
         self._external_connection = False
+        self._cbs_authenticator = None
+        self._auth_timeout = kwargs.pop("auth_timeout", constants.AUTH_TIMEOUT)
```
annatisch, line 124: Maybe this constant should be called `DEFAULT_AUTH_TIMEOUT`

uamqp/client.py
annatisch: We should probably also document what the default value that we're using is.

uamqp/client.py
annatisch: Yeah I think the state check could be moved into `handle_token`

test_perf.py
yunhaoling: Depending on the model, I prefer Credential which represents the data it keeps -- Credential used for authentication, Auth is more like a process.

uamqp/client.py
yunhaoling: code updated ^^

uamqp/authentication.py
yunhaoling: updated!

uamqp/cbs.py
yunhaoling: nope, thanks for pointing out, just removed those ivars

uamqp/client.py
yunhaoling: your understanding is correct, doc updated

uamqp/aio/_link_async.py
yunhaoling: noticed it by an error logging

uamqp/authentication.py
yunhaoling: emmm, wondering what is a "string literal discriminator", a class variable?

annatisch: Yes - exactly that. Just a string class variable:
```python
class Foo:
    kind = 'foo'
    value = 42

class Bar:
    kind = 'bar'
    data = 'some bytes'
```
That way, we can make a cleaner switch statement, rather than type-checking everything:
```python
if obj.kind == 'foo':  # can also use enums here
    output = obj.value
elif obj.kind == 'bar':
    output = obj.data
else:
    raise Exception("what's this")
```

uamqp/authentication.py
```diff
@@ -43,19 +45,22 @@ def _generate_sas_token(auth_uri, sas_name, sas_key, expiry_in=AUTH_DEFAULT_EXPI
 
 
 class SASLPlainAuth(object):
+    _auth_type = AUTH_TYPE_SASL_PLAIN
```
annatisch, line 48: I think we could just make this public. Means we don't have to suppress pylint every time we check it.

annatisch (approved):

PR #249, opened 2021-07-02
create devops CI for pyproto branch

swathipil:


.azure-pipelines/pyproto-ci.yml
```diff
@@ -0,0 +1,70 @@
+trigger:
+  branches:
+    include:
+    - pyproto
```
scbedd, line 4: This is setting filter of `CI` trigger to `pyproto`.

I believe you need to set a `pr` trigger here?

scbedd: [Example](https://github.com/Azure/azure-sdk-for-python/blob/2e3670ae4ed76d994a67ef35901c94f86a1cf5f6/sdk/automanage/ci.yml#L3) of what I mean.

scbedd (approved):
lg2m otherwise!

PR #257, opened 2021-07-09
update client.py with constants

swathipil:
updated for receive event scenario: https://github.com/Azure/azure-sdk-for-python/pull/19748/files

### DIFF
--- a/uamqp/client.py
+++ b/uamqp/client.py
@@ -40,7 +40,6 @@ from .error import AMQPConnectionError
 from .mgmt_operation import MgmtOperation
 from .cbs import CBSAuthenticator
 from .authentication import _CBSAuth
-import pdb
 
 
 _logger = logging.getLogger(__name__)
--- a/uamqp/client.py
+++ b/uamqp/client.py
@@ -22,6 +22,7 @@ from .sender import SenderLink
 from .receiver import ReceiverLink
 from .sasl import SASLTransport
 from .endpoints import Source, Target
+from .error import AMQPConnectionError
 
 from .constants import (
     MessageDeliveryState,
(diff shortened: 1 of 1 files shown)


uamqp/client.py
```diff
@@ -22,6 +22,7 @@
 from .receiver import ReceiverLink
 from .sasl import SASLTransport
 from .endpoints import Source, Target
+from .error import AMQPConnectionError
```
yunhaoling, line 25: is this a duplicate of line 39?

uamqp/client.py
```diff
@@ -291,14 +294,19 @@ def mgmt_request(self, message, operation=None, operation_type=None, node='$mana
...
         status, response, description = mgmt_link.execute(
             message,
             operation=operation,
             operation_type=operation_type,
             timeout=timeout
         )
+        status_code_field = kwargs.pop('status_code_field', b'statusCode')
+        description_field = kwargs.pop('description_field', b'statusDescription')
+        response.application_properties[status_code_field] = status
+        response.application_properties[description_field] = description
```
yunhaoling, line 309: we could remove those first..

I think those field should be handled during mgmt link initialization:
https://github.com/Azure/azure-uamqp-python/blob/pyproto/uamqp/management_link.py#L52-L53

let's revisit this in the work for supporting mgmt link

uamqp/client.py
```diff
@@ -555,9 +562,6 @@ def _message_received(self, message):
             self._message_received_callback(message)
         if not self._streaming_receive:
             self._received_messages.put(message)
-        elif not message.settled:
```
yunhaoling, line 558: how about comment out those lines, and add a TODO: do we need settled property for a message?

uamqp/client.py
swathipil: ah yes it is, good catch

yunhaoling (approved):
LGTM!

PR #260, opened 2021-07-21
[pyproto] Support receiving large messages that are split into multiple transfer frames

yunhaoling:
addressing issue: https://github.com/Azure/azure-sdk-for-python/issues/19869

### DIFF
--- a/uamqp/_transport.py
+++ b/uamqp/_transport.py
@@ -524,7 +524,7 @@ class SSLTransport(_AbstractTransport):
         try:
             while toread:
                 try:
-                    nbytes = self.sock.recv_into(view[nbytes:])
+                    nbytes = self.sock.recv_into(view[length:])
                 except socket.error as exc:
                     # ssl.sock.read may cause a SSLerror without errno
                     # http://bugs.python.org/issue10272
--- a/uamqp/receiver.py
+++ b/uamqp/receiver.py
@@ -74,9 +74,13 @@ class ReceiverLink(Link):
         if not self.received_delivery_id and not self._received_payload:
             pass  # TODO: delivery error
         if self._received_payload or frame[5]:  # more
-            raise NotImplementedError()  # TODO
+            self._received_payload += bytes(frame[11])  # TODO, is it possible to use memoryview directly for concatenation?
         if not frame[5]:
-            message = decode_payload(frame[11])
+            if self._received_payload:
+                message = decode_payload(memoryview(self._received_payload))  # TODO, is it possible to use memoryview directly for concatenation?
+                self._received_payload = b""
+            else:
+                message = decode_payload(frame[11])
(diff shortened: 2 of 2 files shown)


uamqp/receiver.py
```diff
@@ -74,9 +74,13 @@ def _incoming_transfer(self, frame):
         if not self.received_delivery_id and not self._received_payload:
             pass  # TODO: delivery error
         if self._received_payload or frame[5]:  # more
-            raise NotImplementedError()  # TODO
+            self._received_payload += bytes(frame[11])  # TODO, is it possible to use memoryview directly for concatenation?
```
annatisch, line 77: It occurs to me that we need to be clear in our documentation/SDK implementations that a Receiver is not threadsafe....

annatisch: Yeah - I think bytes/memoryview issue might be why I punted this....
You could start with what you have here, to get it functional, leave the "TODOs" in place and then we can revisit this during the perf testing process.

uamqp/_transport.py
```diff
@@ -524,7 +524,7 @@ def _read(self, toread, initial=False, buffer=None,
         try:
             while toread:
                 try:
-                    nbytes = self.sock.recv_into(view[nbytes:])
+                    nbytes = self.sock.recv_into(view[length:])
```
annatisch, line 527: hmm this doesn't break anything?

uamqp/receiver.py
```diff
@@ -74,9 +74,13 @@ def _incoming_transfer(self, frame):
...
         if not frame[5]:
-            message = decode_payload(frame[11])
+            if self._received_payload:
+                message = decode_payload(memoryview(self._received_payload))  # TODO, is it possible to use memoryview directly for concatenation?
+                self._received_payload = b""
```
annatisch, line 81: Are we going to add some tests around this?

uamqp/_transport.py
yunhaoling: nope, it doesn't break -- I would categorize it as a bug fix.

with small frames, nbytes == length, so we're good.

however, with large frames which are split into several reads, we have to move forward the cursor ( which is the `length` ) to properly append the read bytes into the memory view. otherwise we're just inserting the read bytes at the position of `nbytes` which will override bytes in the memoryview.

uamqp/receiver.py
yunhaoling: I have added a test case to receive large and small messages, it runs with no problem

uamqp/receiver.py
yunhaoling: I just put a comment on the client classes docstring mentioning not thread safe, also add an onenote page about the docs.

for receive payload, depending on the send perf PR, I think same bytearray implementation could be applied here, but that would be in a separate PR.

annatisch (approved):

PR #267, opened 2021-08-31
[pyproto] send fixes

yunhaoling:


uamqp/client.py
```diff
@@ -373,12 +373,25 @@ def _transfer_message(self, message_delivery, timeout=0):
...
-            message_delivery.reason = reason
+            # TODO: error action
+            if not state and reason == LinkDeliverySettleReason.NotDelivered:
+                # TODO: message is not delivered
+                message_delivery.state = MessageDeliveryState.Error
+                return
+            error_response = ErrorResponse(state[SEND_DISPOSITION_REJECT])
+            if error_response.condition == b'com.microsoft:server-busy':
+                # TODO: max retries configs
+                time.sleep(4)  # 4 is what we're doing nowadays in EH/SB, service tells client to backoff for 4 seconds
```
annatisch, line 388: We should look at ways of pulling this logic out (serverbusy + backoff) - as I believe it's EH/SB specific and I really don't like how I added it into the uamqp-c version. We can keep it here while we conduct the perf tests, but should look at better designs for this before adding it.

uamqp/client.py
```diff
@@ -373,12 +373,25 @@ def _transfer_message(self, message_delivery, timeout=0):
     def _on_send_complete(self, message_delivery, message, reason, state):
         # TODO: check whether the callback would be called in case of message expiry or link going down
         # and if so handle the state in the callback
-        if SEND_DISPOSITION_ACCEPT in state:
+        if state and SEND_DISPOSITION_ACCEPT in state:
```
annatisch, line 376: In what case will state be None?

uamqp/client.py
```diff
@@ -373,12 +373,25 @@ def _transfer_message(self, message_delivery, timeout=0):
...
             message_delivery.state = MessageDeliveryState.Ok
         else:
             # TODO: sending disposition state could only be rejected/accepted?
-            message_delivery.state = MessageDeliveryState.Error
-            message_delivery.reason = reason
+            # TODO: error action
+            if not state and reason == LinkDeliverySettleReason.NotDelivered:
+                # TODO: message is not delivered
+                message_delivery.state = MessageDeliveryState.Error
```
annatisch, line 383: Shouldn't we add the reason before returning?
`message_delivery.reason = reason`
While 'NotDelivered' is not hugely useful - it's still good to know that it didn't fail _after_ delivery.

uamqp/error.py
```diff
@@ -248,3 +248,25 @@ def __init__(self, status_code, status_description):
         if self.description:
             message += u"\nDescription: {}".format(self.status_description.decode('utf-8'))
         super(TokenAuthFailure, self).__init__(message)
+
+
+class ErrorResponse(object):
+    """
+    """
+    def __init__(self, error_info=None, condition=None, description=None, info=None):
```
annatisch, line 256: All kwargs?

uamqp/error.py
```diff
@@ -248,3 +248,25 @@ def __init__(self, status_code, status_description):
...
+        info = None
+        self.condition = condition
+        self.description = description
+        self.info = info
+        self.error = error_info
+        if isinstance(error_info, list) and len(error_info) >= 1:
+            if isinstance(error_info[0], list) and len(error_info[0]) >= 1:
```
annatisch, line 263: Is error_info completely arbitrary?
i.e. is there actual cases where error_info isn't a list? Or that error_info[0] isn't a list?

uamqp/error.py
```diff
@@ -248,3 +248,25 @@ def __init__(self, status_code, status_description):
...
+                self.condition = error_info[0][0]
+                if len(error_info[0]) >= 2:
+                    self.description = error_info[0][1]
+                if len(error_info[0]) >= 3:
+                    info = error_info[0][2]
+        try:
+            self.info = info.value
```
annatisch, line 270: Will this actually ever have a `.value` attribute? Or is that a symptom of the old c-amqp types?

uamqp/client.py
yunhaoling: In current implementation, it would happen in the following cases when we decide to discard all pending deliveries:

- we initiate detach
- service initiates the detach (a detach frame arrives)
- session state changed to DISCARDING (the connection got CLOSE_RCVD/END)

uamqp/client.py
yunhaoling: updated ^^

uamqp/error.py
yunhaoling: I don't see the AMQP spec has mentioned special about the structure of "state" within a disposition frame -- in the spec, the "state" is of type "*" (Any).

So my understanding is that this is Azure specific, and these errors are conforming to the standard AMQP error structure (condition/description/info).

uamqp/error.py
yunhaoling: updated!

uamqp/error.py
yunhaoling: good catch, I copied the code from uamqp, the `.value` should not be here.

uamqp/client.py
yunhaoling: it's completely on me being too lazy -- I just hacked the code to get it work for perf analysis first.

in uamqp this logic is already pulled out into ErrorPolicy/ErrorAction. I have updated the implementation to reuse our code in uamqp.

I do want la better design for the error handling/retry part, but I think the current uamqp way is be a good start.

annatisch: I'd actually rather we just open an issue to find a design for it than resue the old ErrorPolicy/ErrorAction code.... at least not in its current form. So just hardcoding the error with the sleep with a big bright TODO next to it might be fine.

uamqp/client.py
annatisch: hmm it feels like a bug that state should ever be None.... if this is the only place where we set it to none, maybe with can plug the gap with whatever the initial state is - or does that make the logic more difficult on line 381?

uamqp/error.py
```diff
@@ -248,3 +248,133 @@ def __init__(self, status_code, status_description):
         if self.description:
             message += u"\nDescription: {}".format(self.status_description.decode('utf-8'))
         super(TokenAuthFailure, self).__init__(message)
+
+
+class MessageResponse(AMQPException):
```
annatisch, line 253: I don't think we're using this are we?

uamqp/client.py
yunhaoling: issue logged: https://github.com/Azure/azure-sdk-for-python/issues/20654

uamqp/error.py
yunhaoling: removed first

uamqp/client.py
yunhaoling: according to the definition of spec, the delivery state (within a disposition frame) is used to inform the remote peer of local changes in the state of deliveries according to the spec.

and there're five types of state: `accepted/rejected/released/modified/received`.

according to the definition I think we'd better keep state as None to imply that it is a sender-side error in the case delivery not being sent at all, assigning a state as one of the aforementioned ones makes me feel we're changing the default behavior.

yunhaoling: I have added a TODO

PR #268, opened 2021-09-01
[pyproto] improve send perf

yunhaoling:
fixing issue: https://github.com/Azure/azure-sdk-for-python/issues/20263

to improve sending performance, use bytearray for bytes concatenation instead of raw bytes.

### DIFF
--- a/uamqp/_encode.py
+++ b/uamqp/_encode.py
@@ -329,10 +329,10 @@ def encode_list(output, value, with_constructor=True, use_smallest=True):
         output.extend(ConstructorBytes.list_0)
         return
     encoded_size = 0
-    encoded_values = []
+    encoded_values = bytearray()
     for item in value:
-        encoded_values.append(encode_value(output, item, with_constructor=True))
-        encoded_size += len(encoded_values[-1])
+        encode_value(encoded_values, item, with_constructor=True)
+    encoded_size += len(encoded_values)
     if use_smallest and count <= 255 and encoded_size < 255:
         output.extend(_construct(ConstructorBytes.list_small, with_constructor))
         output.extend(struct.pack('>B', encoded_size + 1))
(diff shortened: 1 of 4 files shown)


uamqp/_encode.py
```diff
@@ -319,28 +326,29 @@ def encode_list(output, value, with_constructor=True, use_smallest=True):
     """
     count = len(value)
     if use_smallest and count == 0:
-        return output + ConstructorBytes.list_0
+        output.extend(ConstructorBytes.list_0)
+        return
     encoded_size = 0
-    encoded_values = []
+    encoded_values = bytearray()
```
annatisch, line 332: This is way nicer :)

uamqp/_encode.py
```diff
@@ -398,37 +406,36 @@ def encode_array(output, value, with_constructor=True, use_smallest=True):
...
     first_item = True
     element_type = None
     for item in value:
         element_type = _check_element_type(item, element_type)
-        encoded_values.append(encode_value(b"", item, with_constructor=first_item, use_smallest=False))
-        encoded_size += len(encoded_values[-1])
+        encode_value(encoded_values, item, with_constructor=first_item, use_smallest=False)
         first_item = False
         if item is None:
             encoded_size -= 1
```
annatisch, line 417: We no longer need this?
In fact in might be messing up the encoded size no longer starting at 0.
Do we even need to declare encoded_size on line 408? Seems we can remove that and just set it on line 419?

yunhaoling: I double check the spec, I feel this is a bug..

an array frame is encoded as the following format:

> array_code - size - count - item_constructor - data1, data2, ...

the size should be the bytes of count + item_constructor + data.
for example, encoding an array which only contains a single None object.

I think the right output bytes should be:
```
b"\xE0\x02\x01\x40"
# 0xE0 indicates a small array
# 0x02 indicates the encoded size which is count + data
# 0x01 indicates the count of the items
# 0x40 is the code for AMQP null
```

PR #271, opened 2021-09-18
Python 3.10 support

yunhaoling:
addressing issue: https://github.com/Azure/azure-sdk-for-python/issues/20622

we're going to drop manylinux1 wheel support for python 3.10.

manyliunx is reaching its EOL on Jan 2022, the pyca team already stops providing cpython 3.10 build-in docker image for manylinux1.

related issue:
https://github.com/pypa/manylinux/issues/994


utils/ManyLinux1_64bit/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM pyca/cryptography-manylinux1:x86_64
+FROM quay.io/pypa/manylinux1_x86_64:latest
```
yunhaoling, line 1: we could drop this dockerfile since we decided not to support manylinux1 for python 3.10

scbedd (approved):
🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀

PR #277, opened 2021-09-29
improve token refresh logic

yunhaoling:
addressing issue: https://github.com/Azure/azure-sdk-for-python/issues/20573

The improvements include:
- allow override refresh window
  - resolve the refresh windows gap between current uamqp setting (6 mins) and AAD service (5 mins)
    - the AAD service returns a cached token, the service side refresh the token until 5 mins before expiration
- not to put an old token onto the cbs link
  - avoid throttling SB/EH service
  - avoid potentially blocking handler work while token is still valid but about to be expired

TODO:

- [x] step1: discuss with anna on the fix/design
  - [x] should override_... be a public parameter?
- [x] step2: add tests

There's one more improvement I could think of, introducing backoff when calling "update_token" if the newly-generated token is the same as old one to avoid throttling token generation module/service, but I'm afraid that goes too far for now.


uamqp/authentication/cbs_auth.py
```diff
@@ -137,7 +139,14 @@ def handle_token(self):
...
                              self._connection.container_id)
                 self.update_token()
-                self._cbs_auth.refresh(self.token, int(self.expires_at))
+                if self.token != self._prev_token:
+                    self._cbs_auth.refresh(self.token, int(self.expires_at))
+                else:
+                    _logger.info(
+                        "The newly acquired token on connection %r is the same as the previous one,"
+                        " will keep attempting to refresh",
+                        self._connection.container_id
```
annatisch, line 148: Is this log statement going to get very noisy? Or is it very rare that token == previous_token after calling `update_token`?

uamqp/authentication/cbs_auth.py
```diff
@@ -231,6 +240,8 @@ def __init__(self, audience, uri, token,
                  **kwargs):  # pylint: disable=no-member
         self._retry_policy = retry_policy
         self._encoding = encoding
+        self._override_token_refresh_window = kwargs.pop("override_token_refresh_window", 0)
```
annatisch, line 243: I don't think we need to call this "override_token_refresh_window" - _all_ the kwargs are overriding _something_.....
Maybe lets just call it 'refresh_window' throughout?

uamqp/authentication/cbs_auth.py
yunhaoling: yes, it would be very noisy in the case `token == previous_token`.

This happens when there's a gap between the client refresh window (uAMQP 6 mins to refresh) and service refresh window (AAD service 5 mins refresh).

So once we have fixed the issue in uamqp/eh/sb, changing default refresh window to 5 mins, it should be rare.

-------------

I think what uamqp is doing technically is not wrong -- it's doing its best to refresh token, however it's doing a bit excessively, retry without backoff. So I think ultimately we might want to introduce kinda backoff to token refresh to let it not throttle

uamqp/authentication/cbs_auth.py
yunhaoling: you're 100% right, suggestion taken!

annatisch (approved):

PR #282, opened 2021-10-18
[pyproto] mgmt request

yunhaoling:
implementing mgmt request
addressing issue: https://github.com/Azure/azure-sdk-for-python/issues/19609

### DIFF
--- a/uamqp/client.py
+++ b/uamqp/client.py
@@ -294,11 +294,7 @@ class AMQPClient(object):
         :keyword str node: The target node. Default node is `$management`.
         :keyword float timeout: Provide an optional timeout in seconds within which a response
          to the management request must be received.
-        :keyword callback: The function to process the returned parameters of the management
-         request including status code and a description if available. This can be used
-         to reformat the response or raise an error based on content. The function must
-         take 3 arguments - status code, response message and description.
-        :paramtype callback: ~callable[int, str, ~uamqp.message.Message]
+        :rtype: ~uamqp.message.Message
         """
 
         # The method also takes "status_code_field" and "status_description_field"
@@ -308,7 +304,6 @@ class AMQPClient(object):
         operation_type = kwargs.pop("operation_type", None)
         node = kwargs.pop("node", "$management")
         timeout = kwargs.pop('timeout', 0)
-        parse_response = kwargs.pop('callback', None)
         try:
             mgmt_link = self._mgmt_links[node]
         except KeyError:
@@ -327,8 +322,6 @@ class AMQPClient(object):
             operation_type=operation_type,
             timeout=timeout
         )
-        if parse_response:
-            return parse_response(status, response, description)
         return response
 
 
--- a/uamqp/client.py
+++ b/uamqp/client.py
@@ -32,7 +32,6 @@ from .constants import (
     SenderSettleMode,
     ReceiverSettleMode,
     LinkDeliverySettleReason,
-    ManagementOpenResult,
     SEND_DISPOSITION_ACCEPT,
     SEND_DISPOSITION_REJECT,
     AUTH_TYPE_CBS,
@@ -317,17 +316,10 @@ class AMQPClient(object):
             mgmt_link = ManagementOperation(self._session, endpoint=node, **kwargs)
             self._mgmt_links[node] = mgmt_link
             mgmt_link.open()
-            while not mgmt_link.mgmt_link_open_status and not mgmt_link.mgmt_error:
+
+            while not mgmt_link.ready():
                 self._connection.listen(wait=False)
-            if mgmt_link.mgmt_error:
-                raise mgmt_link.mgmt_error
-            if mgmt_link.mgmt_link_open_status != ManagementOpenResult.OK:
-                # TODO: update below with correct status code + info
-                raise AMQPConnectionError(
-                    400,
-                    "Failed to open mgmt link: {}".format(mgmt_link.mgmt_link_open_status),
-                    {}
-                )
+
         operation_type = operation_type or b'empty'
         status, description, response = mgmt_link.execute(
             message,
--- a/uamqp/client.py
+++ b/uamqp/client.py
@@ -288,11 +288,11 @@ class AMQPClient(object):
         :type message:
         :keyword str operation:
         :keyword str operation_type:
-        :keyword str node:
+        :keyword str node: Default node is `$management`.
         :keyword float timeout:
         :keyword callback:
-        :keyword status_code_field
-        :keyword status_description_field
+        :keyword status_code_field:
+        :keyword status_description_field:
         :paramtype callback: Callable
         """
         operation = kwargs.pop("operation", None)
--- a/uamqp/client.py
+++ b/uamqp/client.py
@@ -43,7 +43,7 @@ from .constants import (
     MESSAGE_DELIVERY_DONE_STATES,
 )
 
-from .mgmt_operation import MgmtOperation
+from .management_operation import ManagementOperation
 from .cbs import CBSAuthenticator
 from .authentication import _CBSAuth
 
@@ -284,17 +284,26 @@ class AMQPClient(object):
 
     def mgmt_request(self, message, **kwargs):
         """
-        :param message:
-        :type message:
-        :keyword str operation:
-        :keyword str operation_type:
-        :keyword str node: Default node is `$management`.
-        :keyword float timeout:
-        :keyword callback:
-        :keyword status_code_field:
-        :keyword status_description_field:
-        :paramtype callback: Callable
+        :param message: The message to send in the management request.
+        :type message: ~uamqp.message.Message
+        :keyword str operation: The type of operation to be performed. This value will
+         be service-specific, but common values include READ, CREATE and UPDATE.
+         This value will be added as an application property on the message.
+        :keyword str operation_type: The type on which to carry out the operation. This will
+         be specific to the entities of the service. This value will be added as
+         an application property on the message.
+        :keyword str node: The target node. Default node is `$management`.
+        :keyword float timeout: Provide an optional timeout in seconds within which a response
+         to the management request must be received.
+        :keyword callback: The function to process the returned parameters of the management
+         request including status code and a description if available. This can be used
+         to reformat the response or raise an error based on content. The function must
+         take 3 arguments - status code, response message and description.
+        :paramtype callback: ~callable[int, str, ~uamqp.message.Message]
         """
+        # The method also takes "status_code_field" and "status_description_field"
+        # keyword arguments as alternate names for the status code and description
+        # in the response body. Those two keyword arguments are used in Azure services only.
         operation = kwargs.pop("operation", None)
         operation_type = kwargs.pop("operation_type", None)
         node = kwargs.pop("node", "$management")
@@ -303,7 +312,8 @@ class AMQPClient(object):
         try:
             mgmt_link = self._mgmt_links[node]
         except KeyError:
-            mgmt_link = MgmtOperation(self._session, endpoint=node, **kwargs)
+
+            mgmt_link = ManagementOperation(self._session, endpoint=node, **kwargs)
             self._mgmt_links[node] = mgmt_link
             mgmt_link.open()
             while not mgmt_link.mgmt_link_open_status and not mgmt_link.mgmt_error:
@@ -318,7 +328,7 @@ class AMQPClient(object):
                     {}
                 )
         operation_type = operation_type or b'empty'
-        status, response, description = mgmt_link.execute(
+        status, description, response = mgmt_link.execute(
             message,
             operation=operation,
             operation_type=operation_type,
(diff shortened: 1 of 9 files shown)


uamqp/client.py
```diff
@@ -282,11 +282,23 @@ def do_work(self, **kwargs):
             return True
         return self._client_run(**kwargs)
 
-    def mgmt_request(self, message, operation=None, operation_type=None, node='$management', **kwargs):
+    def mgmt_request(self, message, **kwargs):
```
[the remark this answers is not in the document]
annatisch: Yeah - what it's called in the client doesn't bother me much... As I generally consider the entire client class to be legacy only there for the purpose of back compatibility.... so I'm fine either way for the time being.

uamqp/constants.py
```diff
@@ -186,8 +186,6 @@ class ManagementExecuteOperationResult(Enum):
 
     OK = 0
     ERROR = 1
-    FAILED_BAD_STATUS = 2
-    INSTANCE_CLOSED = 3
```
annatisch, line 190: These are no longer used?

uamqp/management_link.py
```diff
@@ -126,27 +128,39 @@ def _on_message_received(self, message):
                 to_remove_operation = operation
                 break
         if to_remove_operation:
-            mgmt_result = ManagementExecuteOperationResult.OK \
-                if 200 <= status_code <= 299 else ManagementExecuteOperationResult.FAILED_BAD_STATUS
             to_remove_operation.on_execute_operation_complete(
-                mgmt_result,
+                ManagementExecuteOperationResult.OK,
```
annatisch, line 132: So if the status code was an error code it would have been raised before this?

uamqp/management_link.py
```diff
@@ -126,27 +128,39 @@ def _on_message_received(self, message):
...
             for operation in self._pending_operations:
                 if message == operation.message:
                     to_remove_operation = operation
                     break
             self._pending_operations.remove(to_remove_operation)
+            # TODO: better error handling
+            error_response = ErrorResponse(error_info=state[SEND_DISPOSITION_REJECT])
+            # The callback is defined in management_operation.py
+            # TODO: AMQPException is too general? to be more specific: MessageReject(Error) or AMQPManagementError?
+            #  or should there an error mapping which maps the condition to the error type
```
annatisch, line 152: With regards to how to handle the error - starting general is fine. 
You will probably find while adapting the EH and SB library that you need finer grain control of errors - then we can usually add more specific subclasses that should be non-breaking.

uamqp/management_link.py
```diff
@@ -126,27 +128,39 @@ def _on_message_received(self, message):
...
             to_remove_operation.on_execute_operation_complete(
-                ManagementExecuteOperationResult.ERROR, None, None, message)
+                ManagementExecuteOperationResult.ERROR,
+                None,
+                None,
+                message,
+                error=AMQPException(
+                    error_response.condition,
+                    error_response.description,
+                    error_response.info,
```
annatisch, line 161: Why do we construct the ErrorResponse object if we're just going to deconstruct it here?

uamqp/management_link.py
```diff
@@ -164,13 +178,23 @@ def execute_operation(
             type=None,
             locales=None
     ):
+        message.application_properties["operation"] = operation
+        message.application_properties["type"] = type
+        message.application_properties["locales"] = locales
+        try:
+            # TODO: namedtuple is immutable, which may push us to re-think about the namedtuple approach for Message
+            new_properties = message.propperties._replace(message_id=self.next_message_id)
```
annatisch, line 186: propperties? :)

uamqp/management_link.py
yunhaoling: nope, in uamqp we do `log.error()` if mgmt_result is not ok.
dealing with the status code is written at upper layer (e.g. service bus).

I will revert this.

uamqp/constants.py
yunhaoling: I was thinking of just letting `ManagementExecuteOperationResult` indicate the result of the mgmt operation itself, e.g. mgmt request succeeds or mgmt request fails and leaving the handling of the status code completely to the upper layer.

But after revisiting the existing uamqp design, I think keeping those status code does no harm, it won't prevent the upper layer handle the raw response messages while it keeps clarity for the response. I will revert the code.

uamqp/management_link.py
yunhaoling: this is a good point, I reused ErrorResponse to parse the error info, but I think I can totally do it with a helper method which should give more flexibility, will do the update

uamqp/management_link.py
yunhaoling: sounds good!

uamqp/management_link.py
yunhaoling: nice catch, updated!

uamqp/client.py
```diff
@@ -282,16 +282,39 @@ def do_work(self, **kwargs):
...
+        :keyword str operation: The type of operation to be performed. This value will
+         be service-specific, but common values include READ, CREATE and UPDATE.
+         This value will be added as an application property on the message.
+        :keyword str operation_type: The type on which to carry out the operation. This will
+         be specific to the entities of the service. This value will be added as
+         an application property on the message.
+        :keyword str node: The target node. Default node is `$management`.
+        :keyword float timeout: Provide an optional timeout in seconds within which a response
+         to the management request must be received.
+        :keyword callback: The function to process the returned parameters of the management
```
annatisch, line 298: Why do we take a callback instead of just returning the response and raising on an error code?

uamqp/management_link.py
```diff
@@ -164,18 +179,36 @@ def execute_operation(
...
+        self._pending_operations.append(PendingManagementOperation(message, on_execute_operation_complete))
 
     def close(self):
         if self.state != ManagementLinkState.IDLE:
             self.state = ManagementLinkState.CLOSING
             self._response_link.detach(close=True)
             self._request_link.detach(close=True)
+            for pending_operation in self._pending_operations:
+                pending_operation.on_execute_operation_complete(
+                    ManagementExecuteOperationResult.INSTANCE_CLOSED,
```
annatisch, line 207: "Instance closed"?
To me "instance" in Python refers to an instance of a class... could maybe just go with 'CLOSED', or 'LINK_CLOSED'

uamqp/management_link.py
```diff
@@ -164,18 +179,36 @@ def execute_operation(
...
+        message = message._replace(properties=new_properties)
 
         self._request_link.send_transfer(
             message,
             on_send_complete=self._on_send_complete,
             timeout=timeout
         )
-        self._pending_operations.append(PendingMgmtOperation(message, on_execute_operation_complete))
+        self.next_message_id += 1
+        self._pending_operations.append(PendingManagementOperation(message, on_execute_operation_complete))
```
annatisch, line 198: Where's is this PendingManagementOperation defined?

uamqp/management_operation.py
```diff
@@ -45,37 +46,63 @@ def _on_amqp_management_open_complete(self, result):
     def _on_amqp_management_error(self):
         """Callback run if an error occurs in the send/receive links."""
         # TODO: This probably shouldn't be ValueError
-        self.mgmt_error = ValueError("Management Operation error ocurred.")
+        self.mgmt_error = ValueError("Management Operation error occurred.")
```
annatisch, line 49: Should this attribute be private?

uamqp/management_operation.py
```diff
@@ -45,37 +46,63 @@ def _on_amqp_management_open_complete(self, result):
...
-        def on_complete(operation_result, status_code, status_description, raw_message, error_condition):
-            if operation_result != ManagementExecuteOperationResult.OK:
-                _LOGGER.error(
-                    "Failed to complete mgmt operation.\nStatus code: %r\nMessage: %r",
-                    status_code, status_description)
-            #message = Message(message=wrapped_message) if wrapped_message else None
-            self._responses[operation_id] = (status_code, raw_message, status_description)
+        self.mgmt_error = None
 
         self._mgmt_link.execute_operation(
```
annatisch, line 87: So `execute_operation` returns once the request has been sent, then the while loop below is used to wait for the response, is that correct?

uamqp/client.py
yunhaoling: This is carried over from [uamqp](https://github.com/Azure/azure-uamqp-python/blob/master/uamqp/client.py#L322) which takes the optional callback to parse the response.

In current service bus, callbacks are defined and passed when calling mgmt request.

I don't have strong preference here, is it that azure-core is doing the way you described? it makes sense to me if we want to align with azure-core implementation.

uamqp/management_link.py
yunhaoling: it's a namedtuple defined at the top of this file
https://github.com/Azure/azure-uamqp-python/blob/pyproto/uamqp/management_link.py#L24

uamqp/management_operation.py
yunhaoling: yep, your understanding is correct

uamqp/management_operation.py
yunhaoling: I have updated the implementation to add a ready() method to `ManagementExecu so that we could make both mgmt_error and mgmt_link_open_status be private

uamqp/management_link.py
yunhaoling: updated name to LINK_CLOSED

uamqp/client.py
annatisch: I'd prefer to remove this callback - I can't see what it actually achieves as it's just being called before returning....
Additionally, the docstring should document the return type - do you know what `response` is that we're returning? Is that just the str?
Seems like we could just be returning a Tuple[int, str].

uamqp/error.py
```diff
@@ -138,6 +137,10 @@ def __init__(self, condition, description, info, message=None):
         self.condition = condition
         self.description = description
         self.info = info
+        message = message or \
+            "Error condition: {}, error description: {}, error info: {}".format(
+              self.condition, self.description, self.info
+            )
         super(AMQPException, self).__init__(message)  # TODO: Pass a message
```
annatisch, line 144: Looks like we can remove the #TODO ;)

uamqp/management_link.py
```diff
@@ -137,16 +143,30 @@ def _on_message_received(self, message):
...
             to_remove_operation = None
             for operation in self._pending_operations:
-                if message == operation.message:
+                if message_delivery.message == operation.message:
                     to_remove_operation = operation
                     break
             self._pending_operations.remove(to_remove_operation)
-            to_remove_operation.on_execute_operation_complete(
-                ManagementExecuteOperationResult.ERROR, None, None, message)
+            if SEND_DISPOSITION_REJECT in state:  # either rejected or accepted
```
annatisch, line 154: We already have this `if` statement on line 147 - so wont it always be true?

uamqp/management_link.py
```diff
@@ -137,16 +143,30 @@ def _on_message_received(self, message):
...
+                #  or should there an error mapping which maps the condition to the error type
+                to_remove_operation.on_execute_operation_complete(
+                    ManagementExecuteOperationResult.ERROR,
+                    None,
+                    None,
+                    message_delivery.message,
+                    error=AMQPException(
+                        state[SEND_DISPOSITION_REJECT][0][0],
+                        state[SEND_DISPOSITION_REJECT][0][1],
+                        state[SEND_DISPOSITION_REJECT][0][2],
```
annatisch, line 167: Nitpick: Can we add a quick # Comment for each line so we can quickly see what each of these indexes represent?

annatisch: Given that we're using Tuples everywhere for performance - I've tried to be super verbose about documenting the indexes with inline-comments to make maintenance easier.

uamqp/management_link.py
```diff
@@ -156,26 +176,52 @@ def open(self):
...
-            operation=None,
-            type=None,
-            locales=None
+        self,
+        message,
+        on_execute_operation_complete,
+        timeout=None,
+        operation=None,
+        type=None,
+        locales=None
```
annatisch, line 185: Is this class public? Should these be keyword-only?

annatisch (approved):

uamqp/client.py
yunhaoling: the response is an amqp message with the body containing the content of the response and app_prop containg response code.

image

uamqp/management_link.py
yunhaoling: nice catch! it's a mistake :P

uamqp/management_link.py
yunhaoling: updated to keyword only

uamqp/management_link.py
yunhaoling: this is a good suggestion, comment added, will keep this in mind

PR #283, opened 2021-10-27
[pyproto] async send and receive

yunhaoling:


uamqp/aio/_cbs_async.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def is_coroutine(get_token):
+    try:
+        if asyncio.iscoroutinefunction(get_token.func):
+            return True
+    except AttributeError:
+        if asyncio.iscoroutinefunction(get_token):
+            return True
+    raise ValueError("get_token must be a coroutine function")
```
annatisch, line 42: If we are enforcing that the token method be async - could we not simply `await` it an let the error raise if they provided a sync function? Rather than checking it?

uamqp/aio/_cbs_async.py
```diff
@@ -0,0 +1,227 @@
...
+        self._token_status_description = None
+
+        self.state = CbsState.CLOSED
+        self.auth_state = CbsAuthState.Idle
+
+    async def _put_token(self, token, token_type, audience, expires_on=None):
+        # type: (str, str, str, datetime) -> None
+        message = Message(
+            value=token,
+            properties=Properties(message_id=self._mgmt_link.next_message_id),
```
annatisch, line 85: Wasn't the management link PR updating the message ID on the outgoing mgmt request?

uamqp/aio/_cbs_async.py
```diff
@@ -0,0 +1,227 @@
...
+            self._on_execute_operation_complete,
+            timeout=self._auth_timeout,
+            operation=CBS_PUT_TOKEN,
+            type=token_type
+        )
+        self._mgmt_link.next_message_id += 1
+
+    async def _on_amqp_management_open_complete(self, management_open_result):
+        if self.state in (CbsState.CLOSED, CbsState.ERROR):
+            _LOGGER.info("Unexpected AMQP management open complete.")
```
annatisch, line 104: This log statement doesn't make sense and would be very confusing :)

uamqp/aio/_cbs_async.py
```diff
@@ -0,0 +1,227 @@
...
+            _LOGGER.info("CBS for connection %r completed opening with status: %r",
+                         self._connection._container_id, management_open_result)
+
+    async def _on_amqp_management_error(self):
+        if self.state == CbsState.CLOSED:
+            _LOGGER.info("Unexpected AMQP error in CLOSED state.")
+        elif self.state == CbsState.OPENING:
+            self.state = CbsState.ERROR
+            await self._mgmt_link.close()
+            _LOGGER.info("CBS for connection %r completed opening with status: %r",
```
annatisch, line 120: Shouldn't this message be "failed to open"?

uamqp/aio/_cbs_async.py
annatisch: The log message is general could probably be a bit more informative as to what these status mean and whether I should do anything. If they are completely non-actionable, then we can probably make them debug.

uamqp/aio/_cbs_async.py
```diff
@@ -0,0 +1,227 @@
...
+                self.state = CbsAuthState.Expired
+            elif is_refresh_required:
+                self.state = CbsAuthState.RefreshRequired
+        elif self.state == CbsAuthState.InProgress:
+            put_timeout = self._check_put_timeout_status()
+            if put_timeout:
+                self.state = CbsAuthState.Timeout
+
+    def _check_expiration_and_refresh_status(self):
+        # TODO: this is duplication the sync version, considering mixin or inheritance
```
annatisch, line 163: Or standalone function? Nothing is being modified on self.

annatisch: Same comment applies to the functions below.

uamqp/aio/_cbs_async.py
```diff
@@ -0,0 +1,227 @@
...
+
+    async def close(self):
+        await self._mgmt_link.close()
+        self.state = CbsState.CLOSED
+
+    async def update_token(self):
+        self.auth_state = CbsAuthState.InProgress
+        access_token = await self._auth.get_token()
+        self._expires_on = access_token.expires_on
+        self._expires_in = self._expires_on - int(utc_now().timestamp())
```
annatisch, line 198: Do we actually need/use _both_ expires_on and expires_in?

uamqp/aio/_client_async.py
```diff
@@ -407,3 +436,111 @@ async def receive_message_batch_async(self, max_batch_size=None, on_message_rece
...
+            if not state and reason == LinkDeliverySettleReason.NotDelivered:
+                message_delivery.state = MessageDeliveryState.Error
+                message_delivery.reason = reason
+                return
+
+            error_response = ErrorResponse(error_info=state[SEND_DISPOSITION_REJECT])
+            # TODO: check error_info structure
+            if error_response.condition == b'com.microsoft:server-busy':
+                # TODO: customized/configurable error handling logic
+                time.sleep(4)  # 4 is what we're doing nowadays in EH/SB, service tells client to backoff for 4 seconds
```
annatisch, line 515: I would rather not have us check for and act on 'microsoft' specific messages....

uamqp/aio/_connection_async.py
```diff
@@ -115,10 +115,11 @@ async def _set_state(self, new_state):
...
         self.state = new_state
-        _LOGGER.info("Connection '%s' state changed: %r -> %r", self.container_id, previous_state, new_state)
-        async with create_task_group() as tg:
-            for session in self.outgoing_endpoints.values():
-                await tg.spawn(session._on_connection_state_change)
+        _LOGGER.info("Connection '%s' state changed: %r -> %r", self._container_id, previous_state, new_state)
+        futures = []
+        for session in self.outgoing_endpoints.values():
+            futures.append(asyncio.create_task(session._on_connection_state_change()))
+        await asyncio.gather(*futures)
```
annatisch, line 122: What happens in the case of an error here? If one session fails to update state, will that raise?

annatisch: What about in the case of an async cancellation?

uamqp/aio/_link_async.py
```diff
@@ -122,9 +121,10 @@ async def _set_state(self, new_state):
             _LOGGER.error("Link state change callback failed: '%r'", e, extra=self.network_trace_params)
 
     async def _remove_pending_deliveries(self):  # TODO: move to sender
-        async with create_task_group() as tg:
-            for delivery in self._pending_deliveries.values():
-                await tg.spawn(delivery.on_settled, LinkDeliverySettleReason.NotDelivered, None)
+        futures = []
+        for delivery in self._pending_deliveries.values():
+            futures.append(asyncio.create_task(delivery.on_settled(LinkDeliverySettleReason.NotDelivered, None)))
+        await asyncio.gather(*futures)
```
annatisch, line 127: Looks like in general we will need a plan for handling the errors when we gather futures like this.

uamqp/aio/_cbs_async.py
yunhaoling: updated

uamqp/aio/_cbs_async.py
yunhaoling: good suggestion! I have created an issue: https://github.com/Azure/azure-sdk-for-python/issues/22835
I believe we need to revisit all the logs in the code

uamqp/aio/_cbs_async.py
yunhaoling: nice suggestion, I was wondering how to avoid duplicate code sync/async module, making them standalone function help resolved the issue.

uamqp/aio/_cbs_async.py
yunhaoling: we only need _expires_on, _expires_in is used to calculate refresh window.

I have updated the code to make _expires_in a local variable

uamqp/aio/_client_async.py
yunhaoling: this part of code has been updated to duplicate the sync implementation

uamqp/aio/_connection_async.py
yunhaoling: I tried the following code and it would raise error `asyncio.exceptions.CancelledError`.

```python
import asyncio

async def func2():
    await asyncio.sleep(5)
    raise ValueError('Error')

async def func3(task):
    await asyncio.sleep(3)
    task.cancel()

async def main():
    task1 = asyncio.create_task(func2())
    task2 = asyncio.create_task(func3(task1))
    await asyncio.gather(task1, task2)

asyncio.run(main())
```

uamqp/aio/_link_async.py
yunhaoling: agreed, this is an interesting problem.

I feel maybe we're good not to deal with the error raised here -- that's what we have in our current sync implementation, but I need to keep thinking more on this and how it would affect developers.

uamqp/aio/_cbs_async.py
yunhaoling: yep, this should be doable, I have removed the coroutine check

uamqp/aio/_cbs_async.py
yunhaoling: I can't remember which PR I wrote, but I have updated the code to duplicate what we have in sync module -- update next_message_id to be +1 after sending out the request, and it works

uamqp/aio/_connection_async.py
annatisch: Interesting that it raised a `CancelledError` - according to the docs:
`the first raised exception is immediately propagated to the task that awaits on gather`
https://docs.python.org/3/library/asyncio-task.html#asyncio.gather
Though either way it looks like errors should be propagated - do we need and error handling for failures in `on_connection_state_changed`?

Lastly - this could be simplified to:
`await asyncio.gather(*[s._on_connection_state_change() for s in self.outgoing_endpoints.values()])`

(List comprehensions will always be more performant than traditional loop syntax ;) )

uamqp/aio/_connection_async.py
```diff
@@ -179,7 +212,18 @@ def _get_next_outgoing_channel(self):
...
+                await self.transport.write(EMPTY_FRAME)
+                self._last_frame_sent_time = time.time()
+        except (OSError, IOError, SSLError, socket.error) as exc:
+            self._error = AMQPConnectionError(
+                ErrorCondition.SocketError,
+                description="Can not send empty frame due to exception: " + str(exc),
+                error=exc
+            )
+        except Exception:
+            raise
```
annatisch, line 226: We don't need this last clause - as that's the implicit default

uamqp/aio/_connection_async.py
```diff
@@ -333,13 +384,26 @@ async def _process_incoming_frame(self, channel, frame):
...
+                    condition=ErrorCondition.ConnectionCloseForced,
+                    description="No frame received for the idle timeout."
+                ),
+                wait=False
+            )
+            return
         await self._send_frame(channel, frame)
 
-    def _get_local_timeout(self, now):
+    async def _get_local_timeout(self, now):
```
annatisch, line 406: Why should this be async?

uamqp/aio/_connection_async.py
```diff
@@ -376,17 +440,30 @@ async def _listen_one_frame(self, **kwargs):
...
             for _ in range(batch):
-                await tg.spawn(self._listen_one_frame, **kwargs)  # TODO: Close on first exception
-
-
-        if self.state not in _CLOSING_STATES:
-            now = time.time()
-            if self._get_local_timeout(now) or (await self._get_remote_timeout(now)):
-                await self.close(error=None, wait=False)
+                futures.append(asyncio.create_task(self._listen_one_frame(**kwargs)))
+            await asyncio.gather(*futures)   # TODO: Close on first exception
```
annatisch, line 453: Another comprehension here:
`await asyncio.gather(*[self._listen_one_frame(**kwargs) for _ in range(batch)])`

uamqp/aio/_session_async.py
```diff
@@ -204,24 +209,75 @@ async def _incoming_flow(self, frame):
...
         else:
-            async with create_task_group() as tg:
-                for link in self._output_handles.values():
-                    if self.remote_incoming_window > 0 and not link._is_closed:
-                        await tg.spawn(link._incoming_flow, frame)
+            futures = []
+            for link in self._output_handles.values():
+                if self.remote_incoming_window > 0 and not link._is_closed:
+                    futures.append(link._incoming_flow(frame))
+            await asyncio.gather(*futures)
```
annatisch, line 216: On another note - we had talked about making Connection/Session/Link objects threadsafe - this will also apply to concurrency-safe.
Wherever we are using `aynscio.gather` - each of those Sessions/Links are behaving concurrently - which could lead to race conditions if they do anything to modify their parent (e.g. if one link errors and updates the session state, a concurrent link might try to set a conflicting session state). 
So we may need to revisit those discussions in this context.

uamqp/aio/_authentication_async.py
```diff
@@ -0,0 +1,76 @@
...
+from ..constants import AUTH_DEFAULT_EXPIRATION_SECONDS
+
+try:
+    from urlparse import urlparse
+    from urllib import quote_plus  # type: ignore
+except ImportError:
+    from urllib.parse import urlparse, quote_plus
+
+
+async def _generate_sas_token_async(auth_uri, sas_name, sas_key, expiry_in=AUTH_DEFAULT_EXPIRATION_SECONDS):
```
annatisch, line 22: We had to make this async?

uamqp/aio/_authentication_async.py
```diff
@@ -0,0 +1,76 @@
...
+
+        """
+        super(SASTokenAuthAsync, self).__init__(
+            uri,
+            audience,
+            username,
+            password,
+            **kwargs
+        )
+        self.get_token = partial(_generate_sas_token_async, uri, username, password, self.expires_in)
```
annatisch, line 76: `partial` supports async right?

uamqp/aio/_cbs_async.py
```diff
@@ -0,0 +1,218 @@
...
+        self._token_put_time = None
+        self._expires_on = None
+        self._token = None
+        self._refresh_window = None
+
+        self._token_status_code = None
+        self._token_status_description = None
+
+        self.state = CbsState.CLOSED
+        self.auth_state = CbsAuthState.Idle
```
annatisch, line 68: Why is CbsState uppscase but not CbsAuthState? :D

uamqp/aio/_client_async.py
```diff
@@ -209,6 +287,176 @@ async def do_work_async(self, **kwargs):
...
+        status, description, response = await mgmt_link.execute(
+            message,
+            operation=operation,
+            operation_type=operation_type,
+            timeout=timeout
+        )
+        return response
+
+
+class SendClient(SendClientSync, AMQPClientAsync):
```
annatisch, line 334: I'd like to scrap this multiple client hierarchy.... though it doesn't need to be done now.

uamqp/aio/_client_async.py
```diff
@@ -209,6 +287,176 @@ async def do_work_async(self, **kwargs):
...
+            MessageDeliveryState.Cancelled,
+            MessageDeliveryState.Timeout
+        ):
+            try:
+                raise message_delivery.error
+            except TypeError:
+                # This is a default handler
+                raise MessageException(condition=ErrorCondition.UnknownError, description="Send failed.")
+
+    async def send_message_async(self, message, **kwargs):
```
annatisch, line 453: We should get rid of all the `_async` suffixes

uamqp/aio/_client_async.py
```diff
@@ -375,35 +675,13 @@ async def receive_message_batch_async(self, max_batch_size=None, on_message_rece
         :param on_message_received: A callback to process messages as they arrive from the
          service. It takes a single argument, a ~uamqp.message.Message object.
         :type on_message_received: callable[~uamqp.message.Message]
-        :param timeout: Timeout in milliseconds for which to wait to receive any messages.
+        :param timeout: I timeout in milliseconds for which to wait to receive any messages.
```
annatisch, line 678: I timeout?

yunhaoling: fixed

uamqp/aio/_cbs_async.py
yunhaoling: nice catch, I updated all the enum to be of upper cases

uamqp/aio/_authentication_async.py
yunhaoling: yes, it supports async

uamqp/aio/_authentication_async.py
yunhaoling: right, this method serves as the async `get_token` method for async auth classes, in which case `get_token` must be a coroutine

uamqp/aio/_connection_async.py
yunhaoling: updated

uamqp/aio/_connection_async.py
yunhaoling: removed

uamqp/aio/_connection_async.py
yunhaoling: nah it doesn't have to -- I was too lazy :P

I have made this a helper method in `connection.py` such that it could be reused in async connection

uamqp/aio/_session_async.py
yunhaoling: yes, thanks for reminding me of the race condition! this is something that we should definitely address..

since currently we are not sharing session/connection, I think we could log a separate issue for race condition and address it in a future PR?

uamqp/aio/_client_async.py
yunhaoling: I also have the same kind of concern for class inheritance -- we use inherence here while not in connection/link/session, I was tempted to align the design.

I would love to create an issue for this and resolve the naming problem (`_async` suffix) in the same issue.

uamqp/aio/_client_async.py
yunhaoling: I think we could resolve the issue at the same time when we resolve the client hierarchy issue.

uamqp/aio/_connection_async.py
yunhaoling: > Though either way it looks like errors should be propagated - do we need and error handling for failures in on_connection_state_changed?

I honestly don't have a clear answer for it.. just taking a look at the code `on_connection_state_changed`, the chain of call is like
1. connection notifies its all subsidiary sessions
2. session notify its all subsidiary links that session state is changed
3. link gets notified and could perform
3.1 attach
3.2 detach and clean up deliveries

I think we should apply single responsibility principle here -- connection module only deals with connection error, error that doesn't belong to this module, leave that error to upstream/downstream to handle..

with that being said, right now, I don't feel we need to handle error here, this is for notifying the upstream that connection state is changed, and upstream modules should take care of themselves.

yunhaoling: (and thanks for the hint of List comprehensions!)

uamqp/aio/_client_async.py
yunhaoling: created an issue: https://github.com/Azure/azure-sdk-for-python/issues/22972

uamqp/aio/_client_async.py
yunhaoling: using the same issue for tracking: https://github.com/Azure/azure-sdk-for-python/issues/22972

uamqp/aio/_session_async.py
yunhaoling: issue created: https://github.com/Azure/azure-sdk-for-python/issues/22973

uamqp/aio/_link_async.py
yunhaoling: issue logged: https://github.com/Azure/azure-sdk-for-python/issues/22975

uamqp/aio/_connection_async.py
annatisch: There are quite a few places (where ever we're currently calling `gather`) that should be replaced with list comprehensions for performance - but we can fix that up later.

annatisch (approved):

PR #287, opened 2021-11-29
Lock.acquire in async connection not awaited

swathipil:
fixes: https://github.com/Azure/azure-sdk-for-python/issues/21639

### DIFF
--- a/uamqp/async_ops/connection_async.py
+++ b/uamqp/async_ops/connection_async.py
@@ -108,7 +108,7 @@ class ConnectionAsync(connection.Connection):
         return self._internal_kwargs.get("loop")
 
     async def lock_async(self, timeout=3.0):
-        await asyncio.wait_for(self._async_lock.acquire(), timeout=timeout, **self._internal_kwargs)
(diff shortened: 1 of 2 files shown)


uamqp/async_ops/connection_async.py
```diff
@@ -108,7 +108,7 @@ def loop(self):
         return self._internal_kwargs.get("loop")
 
     async def lock_async(self, timeout=3.0):
-        await asyncio.wait_for(self._async_lock.acquire(), timeout=timeout, **self._internal_kwargs)
+        await asyncio.wait_for(await self._async_lock.acquire(), timeout=timeout, **self._internal_kwargs)
```
yunhaoling, line 111: I think this would actually give an error?

depending on [the doc of asyncio.wait_for](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for), it's taking an await object.

after adding this await, we turn await object into an unawait object?

swathipil: hmm yeah that's strange, I looked up the usage of .wait_for up and this isn't it. The original should work. I'll keep looking.

PR #289, opened 2021-12-02
Fix thread not started error

Dreamsorcerer:
We keep getting errors from our code that a thread has not started while trying to `.join()` it on line 285.

Most likely the thread failed to start, which throws a `RuntimeError`, then the azure library catches the error and calls the `close()` method. At this point, the thread has been assigned to the attribute, but has failed to start, which causes the code to call `.join()` and error.

Proposed solution is to reset the attribute to `None` when such an error occurs.

### DIFF
--- a/uamqp/client.py
+++ b/uamqp/client.py
@@ -260,11 +260,7 @@ class AMQPClient(object):
             self._build_session()
             if self._keep_alive_interval:
                 self._keep_alive_thread = threading.Thread(target=self._keep_alive)
-                try:
-                    self._keep_alive_thread.start()
-                except RuntimeError:
-                    self._keep_alive_thread = None
-                    raise
+                self._keep_alive_thread.start()
         finally:
             if self._ext_connection:
                 connection.release()
@@ -286,7 +282,10 @@ class AMQPClient(object):
             self.message_handler = None
         self._shutdown = True
         if self._keep_alive_thread:
-            self._keep_alive_thread.join()
+            try:
+                self._keep_alive_thread.join()
+            except RuntimeError:  # Probably thread failed to start in .open()
+                logging.info("Keep alive thread failed to join.", exc_info=True)
             self._keep_alive_thread = None
         if not self._session:
             return  # already closed.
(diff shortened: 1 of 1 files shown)


yunhaoling:
hey @Dreamsorcerer , thanks for the PR, I think your fix should work.

btw, do you know why the thread failed to start -- what's the error information? because depending on the root cause, I think we could choose not to raise the error instead just a warning logging.

Dreamsorcerer:
Either the pymongo and/or azure libraries seem to be using way too many threads, so we seem to keep hitting the limit and the OS won't allow any more threads to be created.

We get these errors in both libraries fairly frequently, so it must be one or the other that is using too much, as both of these libraries are not native asyncio, but async wrappers around a thread-based library.

uamqp/client.py
```diff
@@ -260,7 +260,11 @@ def open(self, connection=None):
             self._build_session()
             if self._keep_alive_interval:
                 self._keep_alive_thread = threading.Thread(target=self._keep_alive)
-                self._keep_alive_thread.start()
+                try:
+                    self._keep_alive_thread.start()
+                except RuntimeError:
+                    self._keep_alive_thread = None
```
yunhaoling, line 266: I personally feel it's simpler to add try/except around `join` in the `close` method instead of mutating the variable on the client during the open stage.

also, through this approach, we could handle other thread-related runtime error unlimited to thread not started.

```python
    def close(self):
        if self._keep_alive_thread:
            try:
                self._keep_alive_thread.join()
            except RuntimeError as exc:
                logging.info("keep alive thread failed to join due to %r", exc)
            finally:
                self._keep_alive_thread = None
```

yunhaoling:
hey @Dreamsorcerer , apologize for missing the thread here.

I agreed with you that this error should not silently pass and should be raised as it will handled by the upper application as well as for the sake of non-breaking behavior.

however, I'm a bit concerned on doing it during open stage, and I think there's a simpler way to do it -- I have left a comment on your PR, please let me know your thoughts on it.

uamqp/client.py
Dreamsorcerer: Fine with me.
Not sure I see any reason for the `finally`, so have skipped that line. Please check changes now.

yunhaoling (approved):
looks good to me! thanks for your contribution!
It will be included in our next release!

PR #290, opened 2021-12-07
[pyproto] error handling and built-in retry strategy

yunhaoling:
Proposal:

- error condition code driven retry mechanism
- strategy for different layers failures:
  - Transport layer: wrapped in AMQPConnection, reestablish every layer and retry
  - AMQP connection: connection close, reestablish every layer and retry
  - AMQP session: session end, reestablish session/link/client and retry
  - AMQP link: link detach, reestablish the link/client and retry
  - AMQP client level: send fail/ receive fail, no need to reestablish, retry the operation
  - Other errors: no close, no retry, just raise(?)
- errors/retry are handled at the client layer
- features need to be retried:
  - send
  - receive
  - mgmt request
- custom error policy for EH and SB, client takes:
  - 1) general retry configs, retry_total/max_backoff/etc. (some of azure-core configs)
  - 2) a map in which key is error condition code, value is the backoff time
  - 3) a list of non retryable error code


uamqp/error.py
```diff
@@ -8,10 +8,133 @@
...
+    UnknownError = b"amqp:unknown-error"
+    VendorError = b"amqp:vendor-error"
+
+
+class RetryMode(str, Enum):
+    Exponential = 'exponential'
+    Fixed = 'fixed'
+
+
+class ErrorPolicy:
```
annatisch, line 51: I think we should try and refactor out the error policy in favour of a retry policy. Where errors are either retryable or they're not.

uamqp/error.py
```diff
@@ -182,7 +319,7 @@ def __init__(self, condition, description=None, info=None):
         super(AMQPConnectionRedirect, self).__init__(condition, description=description, info=info)
 
 
-class AMQPSessionError(AMQPException):
+class AMQPSessionError(AMQPConnectionError):
```
annatisch, line 322: I'm not sure this and the LinkDetach should actually be ConnectionErrors?
I feel like there's a terminology overload here... I thought 'Connection' Error might relate to the Connection layer, but here we're implying that it's related to the network connection?

uamqp/error.py
```diff
@@ -191,16 +328,21 @@ class AMQPSessionError(AMQPException):
     """
 
 
-class AMQPLinkError(AMQPException):
+class LinkDetach(AMQPConnectionError):
     """Details of a Link-level error.
 
-    :param ~uamqp.LinkErrorCondition condition: The error code.
+    :param str condition: The error code.
```
annatisch, line 334: isn't condition bytes?

uamqp/error.py
```diff
@@ -191,16 +328,21 @@ class AMQPSessionError(AMQPException):
     """
 
 
-class AMQPLinkError(AMQPException):
+class LinkDetach(AMQPConnectionError):
```
annatisch, line 331: So we don't need a generic "LinkError" - we only raise errors on Detach?

uamqp/error.py
```diff
@@ -159,6 +288,14 @@ class AMQPConnectionError(AMQPException):
     """
 
 
+class ConnectionClose(AMQPConnectionError):
+    pass
+
+
+class VendorConnectionClose(ConnectionClose):
```
annatisch, line 295: As mentioned above - let's clean out this and VendorLinkDetach

uamqp/link.py
```diff
@@ -108,6 +114,23 @@ def from_incoming_frame(cls, session, handle, frame):
...
+    def _process_link_error(self, condition, description, info):
+        try:
+            amqp_condition = ErrorCodes(condition)
+        except ValueError:
+            error = VendorLinkDetach(condition, description, info)
+        else:
+            if amqp_condition == ErrorCodes.LinkRedirect:
+                error = LinkRedirect(amqp_condition, description, info)
+            else:
+                error = LinkDetach(amqp_condition, description, info)
```
annatisch, line 131: Same comment as connection - I think this function can be replaced with a simple:
```python
self._error = LinkRedirect(...) if condition == LinkRedirect else LinkDetach(...)
```

uamqp/client.py
```diff
@@ -172,6 +180,39 @@ def _client_run(self, **kwargs):
...
+                if absolute_timeout < 0:
+                    raise TimeoutError("Operation timed out.")
+                return operation(*args, timeout=absolute_timeout, **kwargs)
+            except AMQPException as exc:
+                if absolute_timeout >= 0:
+                    retry_active = self._error_policy.increment(retry_settings, exc)
+                    if not retry_active:
+                        break
+                    time.sleep(self._error_policy.get_backoff_time(retry_settings, exc))
+                    if isinstance(exc, LinkDetach):
```
annatisch, line 204: We should avoid instance checking exceptions....

uamqp/_connection.py
```diff
@@ -204,11 +217,18 @@ def _send_frame(self, channel, frame, timeout=None, **kwargs):
...
-            self._transport.send_frame(channel, frame, **kwargs)
+            try:
+                self._last_frame_sent_time = time.time()
+                if timeout:
+                    with self._transport.block_with_timeout(timeout):
+                        self._transport.send_frame(channel, frame, **kwargs)
+                self._transport.send_frame(channel, frame, **kwargs)
+            except Exception as exc:
+                self._error = AMQPConnectionError(
+                    ErrorCodes.InternalError,
```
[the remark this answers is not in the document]
yunhaoling: yes, this should be catching socket errors.

The only error I have encountered is "ConnectionResetError : [winerror 10054] an existing connection was forcibly closed by the remote host" which happened due to 4 mins idle timeout.

Apart from that I see that in our transport.py there're errors like `SSLError` or `socket.error` that could be raised.

Are you talking about introduce a new error type? -- to me I'm treating socket error as kind of AMQPConnectionError..

uamqp/client.py
yunhaoling: I could probably avoid this by checking the error condition 🤔

uamqp/error.py
yunhaoling: is it about having two separate policies like:
```python
class ErrorPolicy:
# in the error policy, we define for each error code we do what
    self.error_map = {
        "some_error_code": (retryable, backoff_time)
    }

class RetryPolicy:
# we perform the retry based on the ErrorPolicy
try:

except Exception as e:
    action = error_policy.get_action(e.condtion)
    # do action

File: uamqp/error.py (hunk 3 above)
Comment (yunhaoling) [reply uamqp/error.py:333]: yes it is, I will update the type

File: uamqp/error.py (hunk 2 above)
Comment (yunhaoling) [reply uamqp/error.py:321]: aha, that's true, multiple "connection" concepts here:

1. the general connection, whenever a client fails to communicate with the service due to amqp link/session/connection or tcp connection error, it's called a connection error.
2. the tcp connection
3. the amqp connection

we're using the same AMQPConnectionError for three different connection concepts..

I do prefer separating them out.

maybe something like the following:

class ConnectionError
class TransportConnectionError(ConnectionError)
class AMQPConnectionError(ConnectionError)
class AMQPSessionError(ConnectionError)
class AMQPLinkError(ConnectionError)

File: uamqp/error.py (hunk 4 above)
Comment (yunhaoling) [reply uamqp/error.py:330]: in general there're two kinds of link level errors:

- 1. errors coming from service which must be via a link detach frame
- 2. errors coming from the client, might be wrong data input, link already closed

to handle errors coming from service, I think LinkDetach is enough (same to Session/Connection as END/CLOSE frame carries the error info).

to handle client side errors, we probably could have this generic "LinkError" as our fallback exception for non-identifies issues and make our error hierarchy extensible.

File: uamqp/_connection.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (annatisch) [reply uamqp/_connection.py:232]: I didn't mean a new error type - rather we should not use the code 'InternalError' - it's a bit ambiguous.

File: uamqp/error.py (hunk 1 above)
Comment (annatisch) [reply uamqp/error.py:50]: hmm this seems a bit complicated - the HTTP retry policy has things like lists of status codes that it acts on etc, which is effectively what we're doing here.

File: uamqp/error.py (hunk 2 above)
Comment (annatisch) [reply uamqp/error.py:321]: This seems a bit much....
Question - if there's an underlying connectivity error/tcp connection failure - what's going to fail first? Will the Session fail first, followed by the AMQP Connection layer? Or will the AMQP Connection layer fail, which in turn will bring down all it's Sessions and Links?

I would probably just start with two: ConnectionError and SessionError.
A ConnectionError indicates the termination of the AMQP Connection, a SessionError indicates the termination of the Session. Both would simply inherit from the base AMQPError.

All of our errors will have a Code attribute to indicate the underlying reason, so we might have something like `ConnectionError.code == SocketFailure` or something.

File: uamqp/error.py (hunk 4 above)
Comment (annatisch) [reply uamqp/error.py:330]: I'd rather not introduce new error types for the sake of it.
We could be consistent with Connection and Session and just name this LinkError rather than LinkDetach. Or we could go the other direction and name the other two ConnectionEnd and SessionClose. In fact, that may not be so bad....

File: uamqp/_connection.py [hunk 9]
```diff
@@ -25,6 +22,12 @@
     ConnectionState
 )
 
+from .error import (
+    ErrorCodes,
+    AMQPConnectionError,
+    ConnectionClose,
+    VendorConnectionClose
```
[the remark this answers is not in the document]
yunhaoling: removed!

uamqp/_connection.py
```diff
@@ -118,6 +122,15 @@ def __enter__(self):
     def __exit__(self, *args):
         self.close()
 
+    def _process_connection_error(self, condition, description, info):
+        try:
+            amqp_condition = ErrorCodes(condition)
+        except ValueError:
+            error = VendorConnectionClose(condition, description, info)
+        else:
+            error = ConnectionClose(amqp_condition, description, info)
```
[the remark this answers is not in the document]
yunhaoling: your suggestion helped simplifies the code a lot -- we just need to create an error and pass all the info when receiving a close frame with error as the following:
```python
if frame[0]:
        self._error = ConnectionClose(
            condition=frame[0][0],
            description=frame[0][1],
            info=frame[0][2]
        )
```

code updated

uamqp/_connection.py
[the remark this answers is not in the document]
yunhaoling: I agree with you.

I was trying to find if there's any shared amqp error condition code we could use, and "internal error" is a general one that could fit -- and that's what uamqp-c does.

honestly, I'm not sure how much freedom we have here in defining condition code. But since this is our own implementation, and the client side will take care of the error, I think probably it's fine to define an "amqp:socket-error" condition?

apart from the condition code itself, I don't think the error handling logic will change as the connection needs to be reestablished anyway.

uamqp/_connection.py
```diff
@@ -204,11 +217,18 @@ def _send_frame(self, channel, frame, timeout=None, **kwargs):
...
+                self._last_frame_sent_time = time.time()
+                if timeout:
+                    with self._transport.block_with_timeout(timeout):
+                        self._transport.send_frame(channel, frame, **kwargs)
+                self._transport.send_frame(channel, frame, **kwargs)
+            except Exception as exc:
+                self._error = AMQPConnectionError(
+                    ErrorCodes.InternalError,
+                    description="Transport error occurred.",
+                    info=exc
```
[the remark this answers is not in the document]
yunhaoling: you're right!

I updated the code based on the idea of core AzureException and updated the AMQPException to take an optional keyword argument "error".

uamqp/_connection.py
```diff
@@ -543,22 +565,34 @@ def listen(self, wait=False, batch=1, **kwargs):
...
-            if self._process_incoming_frame(*new_frame):
-                break
-        if self.state not in _CLOSING_STATES:
-            now = time.time()
-            if self._get_local_timeout(now) or self._get_remote_timeout(now):
-                self.close(error=None, wait=False)
+            for _ in range(batch):
+                new_frame = self._read_frame(wait=wait, **kwargs)
+                if not new_frame:
+                    raise ValueError("Connection closed.")
```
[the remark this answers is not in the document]
yunhaoling: Updated to be a `ConnectionClose` error.

uamqp/_connection.py
```diff
@@ -543,22 +565,34 @@ def listen(self, wait=False, batch=1, **kwargs):
...
+                if self._process_incoming_frame(*new_frame):
+                    break
+            if self.state not in _CLOSING_STATES:
+                now = time.time()
+                if self._get_local_timeout(now) or self._get_remote_timeout(now):
+                    self.close(error=None, wait=False)
+        except Exception as exc:
+            self._error = AMQPConnectionError(
+                ErrorCodes.InternalError,
+                description="Transport error occurred",
```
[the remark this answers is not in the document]
yunhaoling: I don't think shared amqp error condition suffice the goal of clarity here, probably worth considering defining our own condition.

(shared error condition ref: http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-transport-v1.0-os.html#type-amqp-error)

uamqp/error.py
```diff
@@ -8,10 +8,133 @@
 from collections import namedtuple
 import six
 
-from .constants import PORT, FIELD
+from .constants import SECURE_PORT, FIELD
 from .types import AMQPTypes, FieldDefinition
 
 
+class ErrorCodes(Enum):
```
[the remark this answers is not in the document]
yunhaoling: updated to class ErrorCodes(bytes, Enum)

uamqp/error.py
yunhaoling: removed

uamqp/error.py
yunhaoling: are you referring to the [retry settings](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/core/azure-core/azure/core/pipeline/policies/_retry.py#L86) as the ErrorPolicy?

I was initially thinking that `ErrorPolicy` and `RetryPolicy` are the same thing, but in azure-core there's no `ErrorPolicy` just retry policy taking some settings.

We could discuss about it more in tomorrow's sync up -- whether it's more of a design thing/what it should be looking like

uamqp/link.py
yunhaoling: code updated to be like:
```
            error_cls = LinkRedirect if frame[2][0] == ErrorCodes.LinkRedirect else LinkDetach
            self._error = error_cls(condition=frame[2][0], description=frame[2][1], info=frame[2][2])
```

uamqp/_connection.py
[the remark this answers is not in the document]
yunhaoling: I added a `SocketError` and catch socket/os/ssl error in connection, let me know how you like it

uamqp/error.py
yunhaoling: I like your proposal which is simpler and clearer, changed current error hierarchy.

uamqp/error.py
yunhaoling: updated

uamqp/_connection.py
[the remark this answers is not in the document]
yunhaoling: actually, I think here we should not raise a connection error just because we don't receive a frame -- chances are that a receiver keeps listening for frames while the service doesn't have any messages yet and besides we're in non-blocking mode, no frame is received in this iteration doesn't mean the connection is closed.

uamqp/error.py
yunhaoling: I double checked the AMQP spec, I'm on your side now -- I don't want to have a single exception class for each of the condition which seems overkilling.

hierarchy changed back

uamqp/client.py
yunhaoling: updated to use error condition

uamqp/_connection.py
```diff
@@ -201,12 +218,26 @@ def _send_frame(self, channel, frame, timeout=None, **kwargs):
...
-            if timeout:
-                with self._transport.block_with_timeout(timeout):
-                    self._transport.send_frame(channel, frame, **kwargs)
-            self._transport.send_frame(channel, frame, **kwargs)
+            try:
+                self._last_frame_sent_time = time.time()
+                if timeout:
+                    with self._transport.block_with_timeout(timeout):
+                        self._transport.send_frame(channel, frame, **kwargs)
+                self._transport.send_frame(channel, frame, **kwargs)
```
annatisch, line 232: hmm should line 232 be in an `else` statement`?
Otherwise it looks like, if `timeout` is specified, we're going to send twice?

uamqp/_connection.py
```diff
@@ -222,14 +253,29 @@ def _get_next_outgoing_channel(self):
...
             _LOGGER.info("-> empty()", extra=self._network_trace_params)
-        self._transport.write(EMPTY_FRAME)
+        try:
+            raise self._error
+        except TypeError:
+            pass
+        try:
+            if self._can_write():
+                self._transport.write(EMPTY_FRAME)
+                self._last_frame_sent_time = time.time()
```
annatisch, line 270: Maybe we should remove line 260, as it seems to be a duplicate

uamqp/_connection.py
```diff
@@ -350,13 +401,20 @@ def _incoming_close(self, channel, frame):
             return
         if channel > self._channel_max:
             _LOGGER.error("Invalid channel")
```
annatisch, line 403: Looks like we might need to set `self._error` here too?

uamqp/_connection.py
```diff
@@ -475,6 +533,16 @@ def _process_outgoing_frame(self, channel, frame):
             raise ValueError("Connection not configured to allow pipeline send.")
         if self.state not in [ConnectionState.OPEN_PIPE, ConnectionState.OPEN_SENT, ConnectionState.OPENED]:
             raise ValueError("Connection not open.")
+        now = time.time()
+        if self._get_local_timeout(now) or self._get_remote_timeout(now):
+            self.close(
+                error=AMQPError(
+                    condition=ErrorCodes.InternalError.value,
```
annatisch, line 540: Why do we use `.value` here? We didn't on line 348?

uamqp/_connection.py
```diff
@@ -623,16 +715,22 @@ def close(self, error=None, wait=False):
...
+                self._set_state(ConnectionState.CLOSE_PIPE)
+            elif error:
+                self._set_state(ConnectionState.DISCARDING)
+            else:
+                self._set_state(ConnectionState.CLOSE_SENT)
+            self._wait_for_response(wait, ConnectionState.END)
+        except Exception as exc:
+            # If error happened during closing, ignore the error and set state to END
+            _LOGGER.info("An error occurred when closing the connection: %r", exc)
+            self._set_state(ConnectionState.END)
```
annatisch, line 734: Does this need to raise?

uamqp/cbs.py
```diff
@@ -203,10 +204,20 @@ def handle_token(self):
...
+            raise AuthenticationException(
+                condition=ErrorCodes.InternalError,
+                description="Failed to open CBS authentication link."
+            )
         elif self.auth_state == CbsAuthState.Error:
-            raise TokenAuthFailure(self._token_status_code, self._token_status_description, self._encoding)
+            raise TokenAuthFailure(
+                self._token_status_code,
+                self._token_status_description,
+                encoding=self._encoding
```
annatisch, line 215: I'm pretty sure we can strip out all the _encoding stuff at some point.

uamqp/error.py
annatisch: Yup more or less - though in the case of Http retry policy - it's largely a white list of status codes that it's safe to retry on.

uamqp/sender.py
```diff
@@ -149,10 +150,12 @@ def _send_unsent_messages(self):
         self._unsent_messages = unsent
 
     def send_transfer(self, message, **kwargs):
-        if self._is_closed:
-            raise ValueError("Link already closed.")
+        self._check_if_closed()
         if self.state != LinkState.ATTACHED:
-            raise ValueError("Link is not attached.")
+            raise AMQPConnectionError(
```
annatisch, line 155: Should this be a ConnectionError?

uamqp/error.py
```diff
@@ -7,10 +7,133 @@
...
+        **kwargs
+    ):
+        """
+
+        keyword int retry_total:
+        keyword float retry_backoff_factor:
+        keyword float retry_backoff_max:
+        keyword RetryMode retry_mode:
+        keyword list no_retry_condition:
+        keyword dict custom_retry_policy:
```
yunhaoling, line 87: draft: custom_retry_policy should be function,

overload retry_backoff_factor it could either be a float/callable returns a float

condition a,b,c,
condition d,e,f microsoft codes, 4 seconds

input: it's taking the AMQPException
output: float

uamqp/error.py
```diff
@@ -7,10 +7,133 @@
...
+        """
+        self.total_retries = kwargs.pop('retry_total', 3)
+        self.backoff_factor = kwargs.pop('retry_backoff_factor', 0.8)
+        self.backoff_max = kwargs.pop('retry_backoff_max', 120)
+        self.retry_mode = kwargs.pop('retry_mode', RetryMode.Exponential)
+        self.custom_no_retry = self.no_retry + kwargs.pop("no_retry_condition", [])
+        self.custom_retry_policy = kwargs.pop("custom_retry_policy", None)
```
yunhaoling, line 94: if we want to keep: rename like condition backoff

uamqp/error.py
```diff
@@ -7,10 +7,133 @@
...
+    SessionUnattachedHandle = b"amqp:session:unattached-handle"
+    LinkRedirect = b"amqp:link:redirect"
+    LinkStolen = b"amqp:link:stolen"
+    LinkDetachForced = b"amqp:link:detach-forced"
+    LinkTransferLimitExceeded = b"amqp:link:transfer-limit-exceeded"
+    LinkMessageSizeExceeded = b"amqp:link:message-size-exceeded"
+    ClientError = b"amqp:client-error"
+    UnknownError = b"amqp:unknown-error"
+    VendorError = b"amqp:vendor-error"
+    SocketError = b"amqp:socket-error"
```
yunhaoling, line 43: to do: check whether error codes are exposed on EH

uamqp/error.py
```diff
@@ -7,10 +7,133 @@
...
+    SessionHandleInUse = b"amqp:session:handle-in-use"
+    SessionUnattachedHandle = b"amqp:session:unattached-handle"
+    LinkRedirect = b"amqp:link:redirect"
+    LinkStolen = b"amqp:link:stolen"
+    LinkDetachForced = b"amqp:link:detach-forced"
+    LinkTransferLimitExceeded = b"amqp:link:transfer-limit-exceeded"
+    LinkMessageSizeExceeded = b"amqp:link:message-size-exceeded"
+    ClientError = b"amqp:client-error"
+    UnknownError = b"amqp:unknown-error"
+    VendorError = b"amqp:vendor-error"
```
yunhaoling, line 42: we could drop off this one

uamqp/_connection.py
yunhaoling: right, there should be an "else" statement.
it didn't run into error previously because we didn't use "timeout".

uamqp/_connection.py
yunhaoling: good catch

uamqp/sender.py
yunhaoling: ConnectionError doesn't exist anymore.. since it's a sender, I would go with an AMQPLinkError first, also left a TODO comment

uamqp/cbs.py
yunhaoling: left a todo comment

uamqp/_connection.py
yunhaoling: I think the error could be safely swallowed when we're trying to close the connection.
otherwise, we're like keeping the connection from closing itself because of error...

uamqp/_connection.py
yunhaoling: nice catch, it should not be using `.value` since extensible enum is used now.

uamqp/_connection.py
yunhaoling: I double checked the uamqp c logic finding out that it will close with error when meeting with this situation.
I duplicate the logic here.

uamqp/error.py
```diff
@@ -7,10 +7,138 @@
...
+        keyword dict custom_retry_policy:
+        """
+        self.total_retries = kwargs.pop('retry_total', 3)
+        # TODO: A. consider letting retry_backoff_factor be either a float or a callback obj which returns a float
+        #  to give more extensibility on customization of retry backoff time, the callback could take the exception
+        #  as input.
+        self.backoff_factor = kwargs.pop('retry_backoff_factor', 0.8)
+        self.backoff_max = kwargs.pop('retry_backoff_max', 120)
+        self.retry_mode = kwargs.pop('retry_mode', RetryMode.Exponential)
+        self.custom_no_retry = self.no_retry + kwargs.pop("no_retry_condition", [])
```
annatisch, line 97: Rather than make a `custom_no_retry` list, we can just have a parameter called `no_retry`, then:
```
self.no_retry.extend(kwargs.get('no_retry', []))
```

uamqp/error.py
```diff
@@ -7,10 +7,138 @@
 from enum import Enum
 from collections import namedtuple
 
-from .constants import PORT, FIELD
+from .constants import SECURE_PORT, FIELD
 from .types import AMQPTypes, FieldDefinition
 
 
+class ErrorCodes(bytes, Enum):
```
annatisch, line 14: Question - what's the difference between this an the `ErrorCondition` enum on line 142? Can we consolidate them?

uamqp/error.py
yunhaoling: nice suggestion, updated!

uamqp/error.py
yunhaoling: yes we can, the `ErrorCondition` is not used so we could remove it, as well as the Connection/Session/LinkCondition

annatisch: I think I prefer the name `ErrorCondition` to `ErrorCode`.
Also - we should keep the documentation.

yunhaoling: updated!

annatisch (approved):

PR #295, opened 2021-12-16
[pyproto] Filling EventHub live test gaps

yunhaoling:
status:

- [x] EH sync receive livetest
- [x] EH sync send livetest

for complete EH PR, please refer to:
https://github.com/Azure/azure-sdk-for-python/pull/22204/files


uamqp/_decode.py
```diff
@@ -261,12 +264,15 @@ def decode_payload(buffer):
...
+            try:
+                message["sequence"].append(value)
+            except KeyError:
+                message["sequence"] = [value]
         elif descriptor == 119:
             message["value"] = value
         elif descriptor == 120:
             message["footer"] = value
-    return message
+    return Message(**message)
```
annatisch, line 275: I had avoided this (along with `Properties` and `Headers` above) in decode - Not sure it provides much value for the cost. Shouldn't be needed for any of the SDKs.

uamqp/utils.py
```diff
@@ -93,7 +94,44 @@ def add_batch(batch, message):
...
+def encode_str(data, encoding='utf-8'):
+    try:
+        return data.encode(encoding)
+    except AttributeError:
+        return data
+
+
+def normalized_data_body(data, **kwargs):
+    encoding = kwargs.get("encoding", "utf-8")
+    if isinstance(data, list):
```
annatisch, line 106: We shouldn't need these instance checks - where are these utilities called?

uamqp/utils.py
```diff
@@ -93,7 +94,44 @@ def add_batch(batch, message):
     batch.data.append(output)
 
 
+def encode_str(data, encoding='utf-8'):
```
annatisch, line 97: Not sure we need this if we're no longer supporting 2.7

uamqp/utils.py
```diff
@@ -93,7 +94,44 @@ def add_batch(batch, message):
...
+        return [sequence]
+
+
 def get_message_encoded_size(message):
     output = bytearray()
     encode_payload(output, message)
     return len(output)
+
+
+# TODO: to have separate module/class to host those helper AMQP VALUE wrapper method
```
annatisch, line 125: Wrapping things like this in methods is expensive. Method calls in Python are not performant. So we'll need to look at the balance between readable/maintainable code - but also not putting extra method calls along the critical perf path.

uamqp/management_link.py
```diff
@@ -198,7 +198,7 @@ def execute_operation(
         :paramtype type: bytes or str
         :keyword str locales: A list of locales that the sending peer permits for incoming
          informational text in response messages.
-        :keyword float timeout: Provide an optional timeout in milliseconds within which a response
+        :keyword float timeout: Provide an optional timeout in seconds within which a response
```
annatisch, line 201: We don't need to convert the value here? Were we already treating it as seconds?

uamqp/_decode.py
yunhaoling: I agree, it's not necessary as it's private.
I just feel that it's a asymmetry in sending and receiving + type hint (ivar vs a key in a dict), so made the change.

did you avoid this for perf concern?

uamqp/utils.py
yunhaoling: there could be cases that the AMQP spec is explicitly asking for type info like ulong/uint/long/int if the value is an integer.

however, python only have int. so we need to pass extra type information for the correct encoding of an integer to tell whether it's a AMQP long, AMQP short, AMQP int etc.
I think for most cases, default Python-AMQP type mapping is good, but we need a way to override the default type.
right now those utils methods are called by the eventhub sdk.

uamqp/utils.py
yunhaoling: it's called by the eventhub sdk to normalize data body

uamqp/utils.py
yunhaoling: This is only called by `normalized_data_body` to encode string to bytes as data body only takes bytes or list of bytes.

PR #297, opened 2021-12-17
[pyproto] fix decode described value

yunhaoling:


uamqp/_decode.py
```diff
@@ -228,10 +228,11 @@ def _decode_array_large(buffer):
 
 def _decode_described(buffer):
     # type: (memoryview) -> Tuple[memoryview, Any]
-    composite_type = buffer[1]
-    buffer, value = _DECODE_BY_CONSTRUCTOR[buffer[2]](buffer[3:])
+    composite_type = buffer[0]
+    buffer, descriptor = _DECODE_BY_CONSTRUCTOR[composite_type](buffer[1:])
```
annatisch, line 232: Why do we need to decode this? Shouldn't the descriptor constructor always be `buffer[1]` as above?
It seems the value of the descriptor is never used, only the value of the described - so we can be more efficient by simply not decoding it? Or do we need to in order to move the buffer to the correct index?

yunhaoling: it's the second one -- move the buffer to the correct index.

image


as per the spec:
> The descriptor portion of a described format code is itself any valid AMQP encoded value, including
other described values. The formal BNF for constructors is given below.

Parsing the descriptor will move the cursor to the right position of the value.

annatisch: Sounds good - it should be too big a perf issue as I don't think there are too many described values per frame....
We could add a TODO to see if there's a way that we can extract only the size of the descriptor in order to bump the buffer without decoding the value itself.

yunhaoling: I added a todo in the code!

annatisch (approved):

PR #298, opened 2021-12-18
[pyproto] receive should return as soon as there are messages

yunhaoling:


uamqp/client.py
```diff
@@ -640,13 +640,24 @@ def receive_message_batch(self, max_batch_size=None, on_message_received=None, t
             return batch
 
         while receiving and not expired and len(batch) < max_batch_size:
-            receiving = self.do_work(batch=max_batch_size)
+
```
annatisch, line 643: Is this new section of code migrated over from v1?
Seems we would also need it in the async version?
We are checking `qsize` a lot - I wonder how expensive this is....

uamqp/client.py
```diff
@@ -640,13 +640,24 @@ def receive_message_batch(self, max_batch_size=None, on_message_received=None, t
...
+                if timeout and time.time() > timeout:
+                    expired = True
+                    break
+
+                before = self._received_messages.qsize()
+                receiving = self.do_work(batch=max_batch_size)
+                received = self._received_messages.qsize() - before
+
+                if self._received_messages.qsize() > 0 and received == 0:
+                    expired = True
```
annatisch, line 654: Why would this condition result in expiry?

uamqp/client.py
```diff
@@ -640,13 +640,24 @@ def receive_message_batch(self, max_batch_size=None, on_message_received=None, t
...
-            receiving = self.do_work(batch=max_batch_size)
+
+            while receiving and self._received_messages.qsize() < max_batch_size:
+                if timeout and time.time() > timeout:
+                    expired = True
+                    break
+
+                before = self._received_messages.qsize()
+                receiving = self.do_work(batch=max_batch_size)
+                received = self._received_messages.qsize() - before
```
annatisch, line 651: I wonder if it would be more efficient to combine the "did we receive anything" logic with the "get_nowait" logic below - so we only need to count/pull from the queue once. Would that work?

uamqp/client.py
yunhaoling: the naming `expired` is actually a bit ambiguous here, I have removed this variable to just use flag `receiving`.

uamqp/client.py
yunhaoling: I have improved the algorithm to call qsize exactly once in each iteration.
let me know how the new algorithm looks to you.

uamqp/client.py
yunhaoling: I think it would work, I have updated the algorithm here to use counter instead of relying on queue/array directly.
let me know it looks you

uamqp/client.py
```diff
@@ -639,14 +638,30 @@ def receive_message_batch(self, max_batch_size=None, on_message_received=None, t
...
-        while receiving and not expired and len(batch) < max_batch_size:
-            receiving = self.do_work(batch=max_batch_size)
-            while len(batch) < max_batch_size:
-                try:
-                    batch.append(self._received_messages.get_nowait())
-                    self._received_messages.task_done()
-                except queue.Empty:
-                    break
+        to_receive_size = max_batch_size - len(batch)
+        before_queue_size = self._received_messages.qsize()
```
annatisch, line 642: Shouldn't `qsize` always be 0 here?
Otherwise the batch would have been returned on line 639?

uamqp/client.py
```diff
@@ -639,14 +638,30 @@ def receive_message_batch(self, max_batch_size=None, on_message_received=None, t
...
+        while receiving and to_receive_size > 0:
             if timeout and time.time() > timeout:
-                expired = True
+                break
+
+            receiving = self.do_work(batch=to_receive_size)
+            cur_queue_size = self._received_messages.qsize()
+            # after do_work, check how many new messages have been received since previous iteration
+            received = cur_queue_size - before_queue_size
+            if to_receive_size < max_batch_size and received == 0:
```
annatisch, line 652: I'm not sure I follow what we're checking with this condition: `if to_receive_size < max_batch_size`

annatisch: Oh actually I think I get - so if the batch was initially empty, `to_receive_size` == `max_batch_size`.
So we just want to avoid breaking on an empty batch. Yup I think this makes sense :)

uamqp/client.py
annatisch: Yeah I think this looks better. Would be interesting to do a `timeit` comparison.

uamqp/client.py
```diff
@@ -639,14 +638,30 @@ def receive_message_batch(self, max_batch_size=None, on_message_received=None, t
...
-                    break
+        to_receive_size = max_batch_size - len(batch)
+        before_queue_size = self._received_messages.qsize()
+
+        while receiving and to_receive_size > 0:
             if timeout and time.time() > timeout:
-                expired = True
+                break
+
+            receiving = self.do_work(batch=to_receive_size)
```
annatisch, line 648: Remind me - what does specifying `batch=` here achieve? Does this mean we will continue polling the socket for _exactly_ `to_receive_size` number of messages? Or 'best effort'?
I'm wondering if we're duplicating logic - for example if `do_work` is already going to return us exactly the number of events we want, or otherwise timeout, then that's also the logic that this function is implementing. But I can't remember exactly what the `batch` flag does....

uamqp/client.py
yunhaoling: the `break` at line 637 will break the while loop + else statement, so line 639 won't be executed when encountering with `queue.Empty` error

uamqp/client.py
yunhaoling: you got it!

uamqp/client.py
yunhaoling: it ended up being called here: https://github.com/Azure/azure-uamqp-python/blob/pyproto/uamqp/_connection.py#L535-L560

The `batch` is for "The number of frames to attempt to read and process before returning" and depending on the implementation, it's doing `best effort` -- will try to consume as much as `batch` frames, but if there's no more frame coming in, the for loop would break.

uamqp/client.py
yunhaoling: I have written a small "timeit" script result to test it.

based on my scripts, the original one is slight faster resulting in ~1.5 seconds, with the change it's ~1.6 seconds. So it's like a roughly ~6% performance loss after adding the extra logic to enable immediate return in receive message batch method.

I've uploaded the timeit script, let me know if you want me to tweak the script a bit further.

uamqp/client.py
annatisch: Cool okay. I feel like we could shuffle this to be more efficient - but this follows the existing logic and it's working so we'll go with it for now :)

annatisch (approved):

PR #302, opened 2021-12-30
[pyproto] unify to use second as unit

yunhaoling:
it turns out that only docs need to be updated, implementations are correctly using seconds as the unit :)

### DIFF
--- a/uamqp/client.py
+++ b/uamqp/client.py
@@ -81,7 +81,7 @@ class AMQPClient(object):
     :type max_frame_size: int
     :param channel_max: Maximum number of Session channels in the Connection.
     :type channel_max: int
-    :param idle_timeout: Timeout in milliseconds after which the Connection will close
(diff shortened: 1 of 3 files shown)


uamqp/client.py
```diff
@@ -81,7 +81,7 @@ class AMQPClient(object):
     :type max_frame_size: int
     :param channel_max: Maximum number of Session channels in the Connection.
     :type channel_max: int
-    :param idle_timeout: Timeout in milliseconds after which the Connection will close
+    :param idle_timeout: Timeout in seconds after which the Connection will close
```
annatisch, line 84: This PR seems very small - we only needed to fix doc strings? We didn't need to add any multipliers to the values?

yunhaoling: I double check the code and find it's just doc strings.
It must be in the event hub code where I keep flipping between milliseconds and seconds that makes me feel dizzy 🤣

annatisch (approved):

PR #310, opened 2022-02-14
add getstate/setstate to BatchMessage for deepcopy

swathipil:
following up on this issue: https://github.com/Azure/azure-sdk-for-python/issues/22529

reason: BatchMessage inherits from Message, so it uses inherited `__getstate__`/`__setstate__`. `__getstate__` calls the `state` property on the object, which Message has but BatchMessage does not. Overriding `__getstate__`/`__setstate__` on BatchMessage.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,11 @@
 Release History
 ===============
 
+1.5.2 (Unreleased)
++++++++++++++++++++
+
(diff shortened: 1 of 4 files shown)


yunhaoling (approved):
LGTM!

please also help bump the version + update changelog, thanks!

PR #311, opened 2022-02-18
[pyproto] async perf improvement

yunhaoling:
addressing issue: https://github.com/Azure/azure-sdk-for-python/issues/21142
- drop asyncio.wait_for in transport
- drop gather in connection listen to prevent creation of excessive amount of coroutine tasks

### DIFF
--- a/uamqp/aio/_connection_async.py
+++ b/uamqp/aio/_connection_async.py
@@ -422,8 +422,7 @@ class Connection(object):
 
     async def _listen_one_frame(self, **kwargs):
         new_frame = await self._read_frame(**kwargs)
-        if await self._process_incoming_frame(*new_frame):
-            raise ValueError("Stop")  # Stop listening
+        return await self._process_incoming_frame(*new_frame)
 
     async def listen(self, wait=False, batch=1, **kwargs):
         try:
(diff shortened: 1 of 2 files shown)


uamqp/aio/_connection_async.py
```diff
@@ -451,11 +452,10 @@ async def listen(self, wait=False, batch=1, **kwargs):
...
                 return
             try:
-                tasks = [asyncio.ensure_future(self._listen_one_frame(**kwargs)) for _ in range(batch)]
-                await asyncio.gather(*tasks)
+                for _ in range(batch):
+                    await asyncio.ensure_future(self._listen_one_frame(**kwargs))
             except ValueError:
-                for task in tasks:
-                    task.cancel()
+                pass
```
annatisch, line 458: Should we be silently passing this?

yunhaoling: I have updated the code to align with the sync implementation (break instead of catching error by the returning value), perf is not impacted.

uamqp/aio/_connection_async.py
```diff
@@ -450,12 +450,9 @@ async def listen(self, wait=False, batch=1, **kwargs):
...
                 )
                 return
-            try:
-                tasks = [asyncio.ensure_future(self._listen_one_frame(**kwargs)) for _ in range(batch)]
-                await asyncio.gather(*tasks)
-            except ValueError:
-                for task in tasks:
-                    task.cancel()
+            for _ in range(batch):
+                if await asyncio.ensure_future(self._listen_one_frame(**kwargs)):
```
annatisch, line 454: What's the benefit in using `ensure_future` if we're going to directly `await` the coroutine?

annatisch (approved):

uamqp/aio/_connection_async.py
yunhaoling: in my experiment, ensure_future actually gives better performance than await..
let me add a todo to further test the diff here.

PR #312, opened 2022-02-18
[pyproto] keep alive support

yunhaoling:
addressing issue: https://github.com/Azure/azure-sdk-for-python/issues/22298

### DIFF
--- a/uamqp/aio/_client_async.py
+++ b/uamqp/aio/_client_async.py
@@ -187,7 +187,7 @@ class AMQPClientAsync(AMQPClientSync):
                     _logger.info("Keeping %r connection alive. %r",
                                  self.__class__.__name__,
                                  self._connection._container_id)
-                    await asyncio.shield(self._connection._get_remote_timeout(current_time))
+                    await self._connection._get_remote_timeout(current_time)
                     start_time = current_time
                 await asyncio.sleep(1)
         except Exception as e:  # pylint: disable=broad-except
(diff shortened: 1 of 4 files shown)


uamqp/aio/_client_async.py
```diff
@@ -176,6 +176,23 @@ async def _do_retryable_operation_async(self, operation, *args, **kwargs):
...
+        start_time = time.time()
+        try:
+            while self._connection and not self._shutdown:
+                current_time = time.time()
+                elapsed_time = (current_time - start_time)
+                if elapsed_time >= interval:
+                    _logger.info("Keeping %r connection alive. %r",
+                                 self.__class__.__name__,
+                                 self._connection._container_id)
+                    await asyncio.shield(self._connection._get_remote_timeout(current_time))
```
annatisch, line 190: I don't think we need to shield this any more?
Now that we're no longer calling into C code - I think simply propagating an Cancelled error is probably fine?
Or was there another reason we were shielding it?

uamqp/client.py
```diff
@@ -256,6 +276,10 @@ def open(self):
                 auth_timeout=self._auth_timeout
             )
             self._cbs_authenticator.open()
+        if self._keep_alive:
+                self._keep_alive_thread = threading.Thread(target=self._keep_alive_worker)
```
annatisch, line 280: Double indented? :)

annatisch (approved):
Looks good - a few minor comments

uamqp/aio/_client_async.py
yunhaoling: nah.. I can't think of reasons to shield it, I copy-pasted from uamqp code.
removed it.

uamqp/client.py
yunhaoling: nice catch, removed

PR #316, opened 2022-03-21
build using windows 2019 image and vs 2019

yunhaoling:
update using windows 2019 because:

- windows 2016 is dead and not available anymore: https://github.com/actions/virtual-environments/issues/4312
- not using windows 2022 because it doesn't support python3.6: https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md

image

update using vs 2019 as windows 2019 only provides vs 2019

### DIFF
--- a/.azure-pipelines/client.test.live.yml
+++ b/.azure-pipelines/client.test.live.yml
@@ -13,7 +13,7 @@ jobs:
   - job: 'Windows'
 
     pool:
-      vmImage: 'vs2017-win2016'
+      vmImage: 'windows-2019'
 
     timeoutInMinutes: 120
 
--- a/setup.py
+++ b/setup.py
@@ -93,7 +93,7 @@ def get_build_env():
 def get_generator_flags():
     flags = ["-G"]
     if is_win:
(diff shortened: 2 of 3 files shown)


.azure-pipelines/client.test.live.yml
```diff
@@ -13,7 +13,7 @@ jobs:
   - job: 'Windows'
 
     pool:
-      vmImage: 'vs2017-win2016'
+      vmImage: 'windows-2019'
```
yunhaoling, line 16: windows-2022 has dropped python 3.6, since we want to support python 3.6 for now, let's use windows-2019

setup.py
```diff
@@ -93,7 +93,7 @@ def get_build_env():
 def get_generator_flags():
     flags = ["-G"]
     if is_win:
-        flags.append("\"Visual Studio 15 2017\"")
+        flags.append("\"Visual Studio 16 2019\"")
```
yunhaoling, line 96: windows 2019 provides vs 2019 instead of vs 2017

lmazuel (approved):

PR #326, opened 2022-06-23
add value/data/seq properties to message

swathipil:
related to: https://github.com/Azure/azure-sdk-for-python/pull/24259

### DIFF
--- a/uamqp/client.py
+++ b/uamqp/client.py
@@ -386,7 +386,7 @@ class AMQPClient(object):
             return False
         return True
 
-    def do_work(self):
+    def do_work(self, **kwargs): # pylint: disable=unused-argument
         """Run a single connection iteration.
         This will return `True` if the connection is still open
         and ready to be used for further work, or `False` if it needs
--- a/uamqp/message.py
+++ b/uamqp/message.py
@@ -235,6 +235,33 @@ class Message(object):
     def delivery_annotations(self, value):
         self._delivery_annotations = value
 
+    @property
+    def data(self):
+        if not self._message or not self._body:
+            return None
+        # pylint: disable=protected-access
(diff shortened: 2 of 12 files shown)


uamqp/message.py
```diff
@@ -235,6 +235,33 @@ def delivery_annotations(self):
...
+    def data(self):
+        if not self._message or not self._body:
+            return None
+        # pylint: disable=protected-access
+        if self._body.type == c_uamqp.MessageBodyType.DataType:
+            return self._body.data
+        return None
+
+    @property
+    def sequence(self):
```
swathipil, line 248: @annatisch - currently this property is returning the same generator each time. After it's been used once, it returns None. Do you know if there's any way to have it return a new generator each time, like `get_data` does?

swathipil: update: this is not being used up, and works when called multiple times.

uamqp/client.py
```diff
@@ -386,7 +386,7 @@ def client_ready(self):
             return False
         return True
 
-    def do_work(self):
+    def do_work(self, **kwargs): # pylint: disable=unused-argument
```
annatisch, line 389: This needs to be added to `ClientAsync.do_work_async`

swathipil: updated

annatisch (approved):

swathipil:
test run: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1782646&view=results

uamqp/message.py
```diff
@@ -250,7 +250,10 @@ def data(self):
 
     @property
     def sequence(self):
-        if not self._message or not self._body:
+        try:
+            if not self._message or not self._body:
```
swathipil, line 254: BatchMessage doesn't have a _message attribute, so will go into except and return None.

PR #335, opened 2022-08-30
Build macOS wheels for x64 and ARM

tonybaloney:
This PR introduces the cibuildwheel project by PyPA as an easier solution for building wheels across many architectures.

It supports, macOS, Linux and Windows and would remove a lot of scripts and templates from this project.

I've just enabled macOS, but in the CI flow you can see it can do Linux and Windows. For Linux it will compile manylinux and manylinux2014 wheels.

The macOS compiler works with ARM/M1 on GitHub Actions.

The addition of "cmake" and "cython" to `pyproject.toml` removes the requirement for a custom docker container with cmake preinstalled.

### DIFF
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -1,0 +1,28 @@
+name: Wheels
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build_wheels:
+    name: Build wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["macos-11"]
+        # os: ["macos-11", ubuntu-20.04, "windows-latest"]
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build wheels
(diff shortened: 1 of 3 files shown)


tonybaloney:
Related: https://stackoverflow.com/questions/73499286/python-error-when-importing-azure-eventhub-on-mac-monterey/73522403#73522403

tonybaloney:
CC @kashifkhan

scbedd:
Thank you for your contribution @tonybaloney ! I personally wasn't aware this product was even a thing. We definitely _should_ use this if we're capable of it. I'll be leveraging it in other repos for certain!

@kashifkhan I don't see any risks with this methodology or PR. I definitely think we should update the existing pipelines to leverage this library to generate the necessary wheels. I know you had concerns about `openssl` support with it, so I leave that up to you.

My recommendation is we accept this PR and do modifications to existing pipes offline to fully "productionize it". As it stands, we can grab the wheels and upload them for any newly covered platform/pyversion combinations.

scbedd (approved):
I don't personally see any risks to accepting this PR. I leave it up to you @kashifkhan  whether or not we merge it.

PR #350, opened 2022-12-06
[uamqp] Remove References to Six

### LINKED ISSUE #348: six usage is redundant
It seems you do no longer support python 2, so I guess six dependency could be removed?

kashifkhan:
closes #348

### DIFF
--- a/build_many_linux.sh
+++ b/build_many_linux.sh
@@ -39,7 +39,7 @@ export IOTHUB_SAS_POLICY=""
 export IOTHUB_SAS_KEY=""
 
 # Test the wheel
-$PYBIN/pip install "certifi>=2017.4.17" "six~=1.0" "enum34>=1.0.4" "pytest" "pylint";
+$PYBIN/pip install "certifi>=2017.4.17" "enum34>=1.0.4" "pytest" "pylint";
 $PYBIN/pip install uamqp --no-index -f /data/wheelhouse;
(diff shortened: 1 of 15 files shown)


src/amqp_string.pyx
```diff
@@ -45,19 +44,18 @@ cdef class AMQPString(StructBase):
...
-        if six.PY3:
-            try:
-                return as_bytes.decode('UTF-8')
-            except UnicodeDecodeError:
-                pass
+        try:
+            return as_bytes.decode('UTF-8')
+        except UnicodeDecodeError:
+            pass
         return str(as_bytes)
```
kashifkhan, line 51: I've kept this return in here in case there is an error with unicode decode. Wasn't sure if it can be removed

annatisch (approved):
LGTM :)

build_many_linux.sh
```diff
@@ -39,7 +39,7 @@ export IOTHUB_SAS_POLICY=""
 export IOTHUB_SAS_KEY=""
 
 # Test the wheel
-$PYBIN/pip install "certifi>=2017.4.17" "six~=1.0" "enum34>=1.0.4" "pytest" "pylint";
+$PYBIN/pip install "certifi>=2017.4.17" "enum34>=1.0.4" "pytest" "pylint";
```
johanste, line 42: What are we using enum34 for?

Issue #355, opened 2023-01-16
Amqp over websocket is not working and leeds to a connection error


MkKiefer:
# Amqp over websocket connection error

**UAMQP Package Version:** 1.6.3   
**Operating System:** Windows / WSL / Docker   
**Python Version:** 3.8, 3.9, 3.10   

## Describe the bug:
Connection to IoTHub failed with transport type:  `amqp.TransportType.AmqpOverWebsocket` at versions > v1.2.13   
Currently using the version: v1.2.13 (the last version where the amqp over ws works for me) and trying to update to v1.6.3
The websocket connection is needed because some firewalls are blocking the plain amqp connection
This behavior is the same on multiple platforms: 
- x86_x64
- Armv7 (Raspberry 3b+)
- Armv8 (Raspberry 4)

The same error ocurred with the async and sync client.

<details>
  <summary>Logg message</summary>
  
  ### Logs
  ```PYTHON
    2023-01-16 10:11:32,134 uamqp.c_uamqp DEBUG    Creating SASL Mechanism
    2023-01-16 10:11:32,138 uamqp.client DEBUG    Opening client connection.
    2023-01-16 10:11:32,138 uamqp        DEBUG    Initializing platform.
    2023-01-16 10:11:32,138 uamqp.connection INFO     Connection b'7475125f-f01a-4a3a-936c-7fdc8c69f607' state changed from <ConnectionState.START: 0> to <ConnectionState.START: 0>
    2023-01-16 10:11:32,140 uamqp.c_uamqp DEBUG    Wrapping value type: <AMQPType.CompositeType: 23>
    2023-01-16 10:11:32,141 uamqp.c_uamqp DEBUG    Wrapping value type: <AMQPType.CompositeType: 23>
    2023-01-16 10:11:32,142 uamqp.sender INFO     Message sender b'sender-link-2aaa5d30-aa2d-4a79-a028-ee3168794609' state changed from <MessageSenderState.Idle: 1> to <MessageSenderState.Opening: 2> on connection: b'7475125f-f01a-4a3a-936c-7fdc8c69f607'
    2023-01-16 10:11:32,224 uamqp.connection INFO     Connection b'7475125f-f01a-4a3a-936c-7fdc8c69f607' state changed from <ConnectionState.START: 0> to <ConnectionState.START: 0>
[...]

PR #356, opened 2023-01-16
default port handling

MkKiefer:
Update default port handling for  (amqp / amap over websocket).
Set the default port for websocket to 443
Thi should fix the bug in #355


kashifkhan:
Thank you for the PR @MkKiefer :)

PR #370, opened 2023-11-08
Py312

kashifkhan:
Update to support py 3.12 release.


pyproject.toml
```diff
@@ -2,7 +2,7 @@
 requires = [
     "setuptools>=42",
     "wheel",
-    "Cython",
+    "Cython<3.0.0",
```
scbedd, line 5: Hmm. I've never used `Cython` before, but it's a valid pin so ok  👍

scbedd (approved):
All CIBW targeting variables are real. CIBW_Platform where appropriate is correct. This looks solid to me. Famous last words on the test runs 😆

pyproject.toml
kashifkhan: yeah had to pin to a version before 3.0, which was a major change. Our code wont compile with 3 and higher

PR #393, opened 2024-08-16
Update uAMQP vendor

kashifkhan:
Pulling upstream changes to vendored uamqp-c. 
Live test run: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4077788&view=results

Need to investigate the following SB SDK test failures:
- `test_queue_complete_message_on_different_receiver`, which was added to the most recent PR of SB but has not been released
- test_sb_client.py::TestServiceBusClient::test_custom_endpoint_connection_verify_exception with: azure.servicebus.exceptions.ServiceBusError: Handler failed: Authorization timeout..

However, these are not failing locally with the wheels installed, so merging the fix.

### DIFF
--- a/.azure-pipelines/test-eh-sb-sdk.yml
+++ b/.azure-pipelines/test-eh-sb-sdk.yml
@@ -36,7 +36,7 @@ steps:
   - script: |
       echo "SDKPath: $(SdkPath)"
       echo "Tests: $(TestResultsPath)"
-      git clone -b swathipil/devtools/log-pipelinecred-usage https://github.com/swathipil/azure-sdk-for-python.git --single-branch --depth 1
+      git clone -b swathipil/sb/allow-not-testing-pyamqp https://github.com/swathipil/azure-sdk-for-python.git --single-branch --depth 1
     displayName: 'Clone azure-sdk-for-python'
   - task: AzurePowerShell@5
     inputs:
--- a/.azure-pipelines/test-eh-sb-sdk.yml
+++ b/.azure-pipelines/test-eh-sb-sdk.yml
@@ -1,26 +1,38 @@
 steps:
-  - pwsh: |
-      $EHSdkPath='azure-sdk-for-python/sdk/eventhub/azure-eventhub'
-      $EHTestResultsPath='eventhub/azure-eventhub/test-results-live.xml'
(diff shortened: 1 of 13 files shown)


.azure-pipelines/test-eh-sb-sdk.yml
```diff
@@ -61,40 +80,58 @@ steps:
...
-      python -m pip install .
-      python -m pip install -r dev_requirements.txt
-      python -m pip list
-      python -m pytest tests -v --doctest-modules $(IgnoreTests) --junitxml=junit/$(TestResultsPath)
-    displayName: 'Run tests'
+  - task: AzurePowerShell@5
+    displayName: Run Tests (AzurePowerShell@5)
     env:
+      SYSTEM_ACCESSTOKEN: $(System.AccessToken)
       AZURE_SUBSCRIPTION_ID: $(azure-subscription-id)
```
benbp, line 87: This should be removed or replaced inline. Is it needed?

swathipil: It's needed in the SB/EH livetests, b/c we create resources using our own fixture/preparers since they're amqp and not http. We have it available here too: https://github.com/Azure/azure-sdk-for-python/blob/61e0fce835ec2640f1c1ef5ac99fc56c118326d7/sdk/eventhub/tests.yml#L29

by inline, do you mean something like `$env:AZURE_SUBSCRIPTION_ID = $account.Id;` below?

swathipil: addressed!

swathipil: looks like it's not actually being set inline? https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4074639&view=logs&j=3264ca38-67bb-54fa-fd91-0bce2e49d546&t=2a681e29-7c4b-5e61-deff-65adf6363d8f&l=11030

swathipil: changing back for now to test credential

swathipil: OOB: the Get-AzContext.* was not being called correctly. Updated, thanks @benbp!

PR #413, opened 2025-05-20
Fix Cmake version

Eyshika:
some deps are still looking for `cmake 2.8` which causes issues installing with `python 3.12` on mac. I have update cmake versions and small fix in receiver file which wasn't letting it build

### DIFF
--- a/src/message_receiver.pyx
+++ b/src/message_receiver.pyx
@@ -120,7 +120,7 @@ cdef class cMessageReceiver(StructBase):
 
 #### Callbacks (context is a MessageReceiver instance)
 
-cdef void on_message_receiver_state_changed(void* context, c_message_receiver.MESSAGE_RECEIVER_STATE_TAG new_state, c_message_receiver.MESSAGE_RECEIVER_STATE_TAG previous_state) noexcept:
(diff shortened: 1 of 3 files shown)


Eyshika:
Please review @scbedd

Eyshika:
@microsoft-github-policy-service agree company="Worlds Enterprise"

scbedd (approved):

scbedd:
@Eyshika . I really appreciate your contribution. I will merge it, but you should know that it is unlikely that it will be released.

[This package is no longer maintained. See here on the root readme for the disclaimer.](https://github.com/Azure/azure-uamqp-python?tab=readme-ov-file#disclaimer)

If other fixes pile up and force us to cut a release, we will, but I think you're probably best served to just build `main` yourself after the merge if you need this package.

Another option would be to crib the [vendored amqp pure-python solution](https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/eventhub/azure-eventhub/azure/eventhub/amqp) that `azure-servicebus` and `azure-eventhubs` use in favor of the platform-specific from this repo.

